PR #66, opened 2023-09-13
Feature/fetch granted access optionnal protected data

PierreJeanjacquot:
this allow to fetch all the granted access (datasetorders) available for a given app or user
:warning: datasets may not be protected data :warning:

### DIFF
--- a/src/dataProtector/fetchGrantedAccess.ts
+++ b/src/dataProtector/fetchGrantedAccess.ts
@@ -6,18 +6,17 @@ import {
 import { WorkflowError } from '../utils/errors.js';
 import {
   addressOrEnsOrAnySchema,
-  addressOrEnsSchema,
   throwIfMissing,
 } from '../utils/validators.js';
 import { formatGrantedAccess } from '../utils/format.js';
 
 export const fetchGrantedAccess = async ({
   iexec = throwIfMissing(),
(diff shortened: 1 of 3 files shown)


luisccastillo (approved):
Approved. Seems good to me but I have a little doubt, when the dataset value is not passed, is it transformed correctly into an empty value ? i.e. 0x0... just to avoid api call errors to the market api. 
For example:
 https://api.market.v8-bellecour.iex.ec/datasetorders?chainId=134&dataset=0x0000000000000000000000000000000000000000 OK  
 
https://api.market.v8-bellecour.iex.ec/datasetorders?chainId=134  NOK

PierreJeanjacquot:
> Approved. Seems good to me but I have a little doubt, when the dataset value is not passed, is it transformed correctly into an empty value ? i.e. 0x0... just to avoid api call errors to the market api. For example: https://api.market.v8-bellecour.iex.ec/datasetorders?chainId=134&dataset=0x0000000000000000000000000000000000000000 OK
> 
> https://api.market.v8-bellecour.iex.ec/datasetorders?chainId=134 NOK

When the dataset is not passed, the value is transformed to `any` and results to calling <https://api.market.v8-bellecour.iex.ec/datasetorders?chainId=134&dataset=any>

PR #81, opened 2023-10-13
Feature/update constructor options

abbesBenayache:


SeddikBellamine (changes requested):
well done! Nice code quality!  just add a test in grant access that uses iexecOptions:{iexecGatewayURL: 'https://custom-market-api-url.com'}, then grant access with this custom marketapi url

SeddikBellamine (approved):

PR #88, opened 2023-11-06
ethers 6 migration

PierreJeanjacquot:


src/dataProtector/protectDataObservable.ts
```diff
@@ -150,8 +149,12 @@ export const protectDataObservable = ({
               e
             );
           });
-        const protectedDataAddress = transactionReceipt.events[1].args[0];
-        const txHash = transactionReceipt.transactionHash;
+
+        const protectedDataAddress = transactionReceipt.logs.find(
+          ({ eventName }) => 'DatasetSchema' === eventName
+        )?.args[0];
```
PierreJeanjacquot, line 155: find the event by name in logs instead of assuming the event is at index [1]

src/utils/validators.ts
```diff
@@ -1,13 +1,11 @@
...
 import { ValidationError, number, object, string, array } from 'yup';
 
-const { isAddress } = utils;
-
 export const throwIfMissing = (): never => {
   throw new ValidationError('Missing parameter');
 };
 
-const isUndefined = (value: any) => value === undefined;
+const isUndefined = (value: unknown) => value === undefined;
```
PierreJeanjacquot, line 8: fix eslint warning

src/dataProtector/protectDataObservable.ts
```diff
@@ -150,8 +154,12 @@ export const protectDataObservable = ({
...
```
PierreJeanjacquot, line 155: find the event by name in logs instead of assuming the event is at index [1]

Le-Caignec (approved):
Why do we want to restrict our tests with wallets using only BIP32 and not BIP39 which include natively BIP32 standard? Moreover, BIP39 is the most commonly use in the ecosystem nowadays.

PierreJeanjacquot:
> Why do we want to restrict our tests with wallets using only BIP32 and not BIP39 which include natively BIP32 standard? Moreover, BIP39 is the most commonly use in the ecosystem nowadays.

@Le-Caignec It's not something we particularly want, signing tx with BIP32 or BIP39 both results to valid signatures so both are suitable. The only thing that make us use BIP32 is that [ethers `Wallet.createRandom()`](https://docs.ethers.org/v6/api/wallet/#Wallet_createRandom) returns a [`HDNodeWallet`](https://docs.ethers.org/v6/api/wallet/#HDNodeWallet) which is a BIP32.

PR #89, opened 2023-11-07
Use swc to transpile TS for tests

cedric25:
- Stop running `tsc build` prior to tests
 - Stop importing from `dist/`


tests/e2e/IExecDataProtector/constructor.test.ts
Le-Caignec, line 1: may be set those comments for the entire file. At the top ?

Le-Caignec (approved):

tests/e2e/IExecDataProtector/constructor.test.ts
cedric25: Your comment made me realize I could just use `graphQLClientUrl['url']` to access and test the private field, instead of these "ignore" comments. Updated 👍 Thanks

PR #95, opened 2023-11-16
setup default page size & page number configuration

SeddikBellamine:
Currently: fetchProtectedData returns the first 1000 results and cuts the next results.

Objectives:

Allow the user to access results beyond the 1000th item

Allow the user to fetch smaller results set

implemented pagination by adding page & pageSize optional parameters to fetchProtectedData . 

10 <= pageSize <= 1000 (default TBD)

0 <= page (default 0)

i setup default pageSize to 1000(TBD)

implemented e2e tests of the fetchProtectedData with pagination.

### DIFF
--- a/src/utils/validators.ts
+++ b/src/utils/validators.ts
@@ -55,8 +55,10 @@ export const positiveIntegerStringSchema = () =>
     '${path} should be a positive integer',
     (value) => isUndefined(value) || isPositiveIntegerStringTest(value)
   );
-export const numberNonNegativeSchema = () =>
-  number().integer().min(0).typeError(`$\{path} must be a non-negative number`);
+
+export const positiveNumberSchema = () =>
+  number().integer().min(0).typeError('${path} must be a non-negative number');
+
 export const numberBetweenSchema = (min: number, max: number) =>
   number()
     .integer()
@@ -87,9 +89,6 @@ export const grantedAccessSchema = () =>
   })
     .noUnknown()
     .default(undefined);
-
-export const positiveNumberSchema = () =>
-  number().integer().min(0).typeError('${path} must be a non-negative number');
 
 export const urlArraySchema = () => array().of(urlSchema());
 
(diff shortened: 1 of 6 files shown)


tests/e2e/IExecDataProtector/fetchProtectedData.test.ts
```diff
@@ -62,4 +62,46 @@ describe('dataProtector.fetchProtectedData()', () => {
...
+    expect(res.length).toBe(1000);
+  });
+  it('pagination: fetches a specific page with a specified page size', async () => {
+    const page = '2'; // Specify the desired page number
+    const pageSize = '50'; // Specify the desired page size
+    const res = await dataProtector.fetchProtectedData({ page, pageSize });
+
+    // Check if the correct number of items for the specified page size is retrieved
+    expect(res.length).toBe(50);
+    // TODO: implement logic to ensure that the items are from the desired page.
```
cedric25, line 76: Maybe we can perform a second request with `page: 0, pageSize: 150` and check that two elements are the same?

Looks like it works:

```
  it('pagination: fetches a specific page with a specified page size', async () => {
    const page = 2; // Specify the desired page number
    const pageSize = 50; // Specify the desired page size
    const res = await dataProtector.fetchProtectedData({ page, pageSize });

    // Check if the correct number of items for the specified page size is retrieved
    expect(res.length).toBe(50);

    const res2ToCheck = await dataProtector.fetchProtectedData({
      page: 0,
      pageSize: 150,
    });
    expect(res[49]).toEqual(res2ToCheck[149]);
  });
```

src/utils/validators.ts
```diff
@@ -11,6 +11,8 @@ const isEnsTest = (value: string) => value.endsWith('.eth') && value.length > 6;
 const isAnyTest = (value: string) => value === 'any';
 
 const isPositiveIntegerStringTest = (value: string) => /^\d+$/.test(value);
+const isBetween10And1000Test = (value: string) =>
+  /^(?!0*$)([1-9]\d{0,2}|1000)$/.test(value);
```
cedric25, line 15: Regex quite hard to read, and it looks like '1' is valid whether it should not?

Capture d’écran 2023-11-21 à 10 26 17

I would suggest:

```
export const numberBetweenSchema = (min: number, max: number) =>
  number()
    .integer()
    .min(min)
    .max(max)
    .typeError(`$\{path} must be a number between ${min} and ${max}`);
```

PR #98, opened 2023-12-04
remove useless test

Le-Caignec:


tests/e2e/IExecDataProtector/fetchProtectedData.test.ts
```diff
@@ -64,11 +64,6 @@ describe('dataProtector.fetchProtectedData()', () => {
     );
   });
 
-  it('pagination: fetches the first 1000 items by default', async () => {
-    const res = await dataProtector.fetchProtectedData();
-    expect(res.length).toBe(1000);
-  });
```
cedric25, line 70: That was to check that by default dataProtector includes a default pageSize = 1000.
Might be worth keeping it? Maybe change its description?

I've also changed it to `.toBeLessThanOrEqual(1000)` in another PR to also pass when there is less than 1,000 protected data.

PR #100, opened 2023-12-05
Also check if instanceof ArrayBuffer when creating zip file

cedric25:


Le-Caignec (approved):
LGTM 🚀 
To see with pierre if we want to accept any data if it's not a : number, boolean, object

PR #103, opened 2023-12-19
specify node version 18.19

SeddikBellamine:
specify node version 18.19 in drone ci


Le-Caignec (changes requested):
You miss one line 203

PR #107, opened 2023-12-21
Upate data unit test to cover ArrayBuffer types

luisccastillo:
Updating utils.data to cover the correction on https://github.com/iExecBlockchainComputing/dataprotector-sdk/commit/6e9c0ea30d0448a80f5d5f14cb676b0cd166cca2

### DIFF
--- a/tests/unit/utils/data.test.ts
+++ b/tests/unit/utils/data.test.ts
@@ -17,7 +17,10 @@ const uint8ArraysAreEqual = (a: Uint8Array, b: Uint8Array) => {
   return a.every((val, i) => val === b[i]);
 };
 
-const uint8ArrayAndArrayBufferAreEqual = (uint8Array: Uint8Array, buffer: ArrayBuffer): boolean => {
+const uint8ArrayAndArrayBufferAreEqual = (
+  uint8Array: Uint8Array,
+  buffer: ArrayBuffer
+): boolean => {
   const bufferUint8Array = new Uint8Array(buffer);
   if (bufferUint8Array.length !== uint8Array.length) {
     return false;
@@ -28,7 +31,7 @@ const uint8ArrayAndArrayBufferAreEqual = (uint8Array: Uint8Array, buffer: ArrayB
     }
   }
   return true;
-}
+};
 
 let data: any;
 let pngImage: Uint8Array;
@@ -69,7 +72,7 @@ beforeEach(async () => {
               pngImage,
               svgImage,
               pdfFile,
-              mp3AudioFile
+              mp3AudioFile,
             },
           },
         },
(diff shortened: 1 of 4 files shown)


tests/unit/utils/data.test.ts
```diff
@@ -311,6 +339,7 @@ describe('createZipFromObject()', () => {
 
   it('serializes binary data as binary file', async () => {
     const zipFile = await createZipFromObject(data);
```
PierreJeanjacquot, line 341: as we are testing a bug fix for binary data loaded as ArrayBuffer we may want to ensure the test data are the type we expected (in this case what was declared as ArrayBuffer was not ArrayBuffer but a nodejs Buffer also instanceof Uint8Array)

the suggestion bellow add checks on data type

```suggestion
    // ensure some data are Uint8Array (nodejs)
    expect(
      data.nested.object.with.binary.data.pngImage instanceof ArrayBuffer
    ).toBe(false);
    expect(
      data.nested.object.with.binary.data.pngImage instanceof Uint8Array
    ).toBe(true);
    // ensure some data are ArrayBuffer (web)
    expect(
      data.nested.object.with.binary.data.pdfFile instanceof ArrayBuffer
    ).toBe(true);
    expect(
      data.nested.object.with.binary.data.pdfFile instanceof Uint8Array
    ).toBe(false);
    const zipFile = await createZipFromObject(data);
```

tests/unit/utils/data.test.ts
```diff
@@ -17,9 +17,27 @@ const uint8ArraysAreEqual = (a: Uint8Array, b: Uint8Array) => {
...
+  }
+  for (let i = 0; i < bufferUint8Array.length; i++) {
+    if (bufferUint8Array[i] !== uint8Array[i]) {
+      return false;
+    }
+  }
+  return true;
+};
+
 let data: any;
```
PierreJeanjacquot, line 36: node `fs/promises` `readFile()` returns a `Buffer`, for the sake of testing with `ArrayBuffer` we need to convert node's `Buffer` to an `ArrayBuffer`

```suggestion
const nodeBufferToArrayBuffer = (nodeBuffer: Buffer): ArrayBuffer =>
  nodeBuffer.buffer.slice(
    nodeBuffer.byteOffset,
    nodeBuffer.byteLength + nodeBuffer.byteOffset
  );


let data: any;
```

tests/unit/utils/data.test.ts
```diff
@@ -29,6 +47,12 @@ beforeAll(async () => {
   svgImage = await fsPromises.readFile(
     path.join(process.cwd(), 'tests', '_test_inputs_', 'image.svg')
   );
+  pdfFile = await fsPromises.readFile(
+    path.join(process.cwd(), 'tests', '_test_inputs_', 'application.pdf')
+  );
+  mp3AudioFile = await fsPromises.readFile(
+    path.join(process.cwd(), 'tests', '_test_inputs_', 'audio.mp3')
+  );
```
PierreJeanjacquot, line 55: convert returned `Buffer` to `ArrayBuffer`

```suggestion
  pdfFile = nodeBufferToArrayBuffer(
    await fsPromises.readFile(
      path.join(process.cwd(), 'tests', '_test_inputs_', 'application.pdf')
    )
  );
  mp3AudioFile = nodeBufferToArrayBuffer(
    await fsPromises.readFile(
      path.join(process.cwd(), 'tests', '_test_inputs_', 'audio.mp3')
    )
  );
```

PierreJeanjacquot (changes requested):
The test was actually not testing the fix 6e9c0ea30d0448a80f5d5f14cb676b0cd166cca2
`git revert 6e9c0ea30d0448a80f5d5f14cb676b0cd166cca2` and `npm run test unit/utils/data` to check the test pass without the fix)

I added some suggestions to enforce using `ArrayBuffer` in the test

PierreJeanjacquot (approved):
LGTM

PR #112, opened 2024-01-10
Some fixes around protectedData schema

cedric25:


src/utils/data.ts
```diff
@@ -114,8 +114,17 @@ export const extractDataSchema = async (
...
+        let guessedTypes: Array<{
+          mime?: string;
+          extension?: string;
+          typename: string;
+        }> = [];
+        if (value instanceof Uint8Array) {
+          guessedTypes = filetypeinfo(value);
+        } else {
+          guessedTypes = filetypeinfo(new Uint8Array(value));
+        }
```
cedric25, line 127: Take `ArrayBuffer` into account:
Have a schema like `file: "application/octet-stream"` instead of `file: {}`

src/utils/data.ts
```diff
@@ -203,8 +212,11 @@ export const createZipFromObject = (obj: unknown): Promise<Uint8Array> => {
 };
 
 export const reverseSafeSchema = function (
-  schema: Array<Record<'id', string>>
+  schema?: Array<Record<'id', string>>
 ) {
+  if (!schema) {
+    return {};
+  }
```
cedric25, line 219: Should never happen but be safe.

src/dataProtector/fetchProtectedData.ts
```diff
@@ -75,7 +75,9 @@ export const fetchProtectedData = async ({
         owner {
           id
         }
-        schema
+        schema {
+          id
+        }
```
cedric25, line 80: This prop was invalid so it was ignored. Fix it.

tests/unit/utils/data.test.ts
```diff
@@ -264,6 +264,14 @@ describe('extractDataSchema()', () => {
     const dataSchema: any = await extractDataSchema(data);
     expect(dataSchema.unknown).toBe('application/octet-stream');
   });
+
+  it('Text-based files also considered as octet-stream for now', async () => {
+    data.text = await fsPromises.readFile(
+      path.join(process.cwd(), 'tests', '_test_inputs_', 'text.txt')
+    );
+    const dataSchema: any = await extractDataSchema(data);
```
Le-Caignec, line 272: we don't want that dataschema look something like that no ? 
```{
        numberZero: 'number',
        numberOne: 'number',
        numberMinusOne: 'number',
        booleanTrue: 'boolean',
        booleanFalse: 'boolean',
        string: 'string',
        nested: { object: { with: [Object] } },
        text: 'application/octet-stream'
      }

Le-Caignec (commented):
some questions

Le-Caignec (approved):
LGTM

PR #117, opened 2024-01-12
Feature/v2 monorepo architecture

Le-Caignec:
- Just copy existing repository : sharing-smart-contract
- move some repo into new folder packages (dapp + deployment-dapp)

TO DO :
- Adapt drone CI (target,....)
- Make tests on drone


Le-Caignec:
We should discuss with the team where `deployment-dapp` folder should go

PR #121, opened 2024-01-16
createCollection() SDK method

cedric25:


packages/sdk/tests/e2e/sharing/createCollection.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  describe('When calling createCollection()', () => {
+    it(
+      'should answer with a collection address',
+      async () => {
+        const { collectionId } = await dataProtector.createCollection();
+        expect(collectionId).toBeDefined();
+      },
+      2 * MAX_EXPECTED_BLOCKTIME
```
cedric25, line 22: Check if good timeout value here.

packages/sdk/src/dataProtector/types.ts
```diff
@@ -348,6 +348,9 @@ export type TransferResponse = {
   txHash: string;
 };
 
+// Sharing methods
+export type { CreateCollectionResponse } from "./sharing/createCollection.js"
```
cedric25, line 352: Trying something here to slow down this file length: the method file will own its types, just re-exporting them here.

packages/sdk/src/dataProtector/sharing/createCollection.ts
```diff
@@ -0,0 +1,42 @@
+import { ethers, type Contract } from 'ethers';
+import { ABI as sharingABI } from '../../contracts/sharingAbi.js';
+import { WorkflowError } from '../../utils/errors.js';
+import { throwIfMissing } from '../../utils/validators.js';
+import { AddressOrENS, IExecConsumer } from '../types.js';
+
+export type CreateCollectionResponse = {
+  collectionId: AddressOrENS;
+};
```
cedric25, line 9: Trying something here to slow down `types.ts` file length: the method file will own its types as we can see here.

packages/sdk/src/dataProtector/sharing/createCollection.ts
```diff
@@ -0,0 +1,42 @@
+import { ethers, type Contract } from 'ethers';
+import { ABI as sharingABI } from '../../contracts/sharingAbi.js';
+import { WorkflowError } from '../../utils/errors.js';
+import { throwIfMissing } from '../../utils/validators.js';
+import { AddressOrENS, IExecConsumer } from '../types.js';
+
+export type CreateCollectionResponse = {
+  collectionId: AddressOrENS;
```
PierreJeanjacquot, line 8: collectionId will be a number

cedric25: Absolutely ✅

packages/sdk/src/dataProtector/sharing/createCollection.ts
```diff
@@ -0,0 +1,42 @@
...
+      );
+    });
+
+  const mintedTokenId = transactionReceipt.logs.find(
+    ({ eventName }) => eventName === 'Transfer'
+  )?.args[2];
+
+  return {
+    collectionId: mintedTokenId,
+  };
```
PierreJeanjacquot, line 41: As tokenid is an uint256 in the ABI `mintedTokenId` is going to be a bigint. Assuming we are using a sequence for collectionIds, we can safely cast it into a number that will fit into safe integers.

```suggestion
  const mintedTokenId = transactionReceipt.logs.find(
    ({ eventName }) => eventName === 'Transfer'
  )?.args[2];

  return {
    collectionId: Number(mintedTokenId),
  };
```

PierreJeanjacquot (changes requested):
good job, see the comments for improvement suggestions

packages/sdk/src/dataProtector/sharing/createCollection.ts
cedric25: Would it make sense to add `as bigint` line 37 to better understand the Number casting line 40?

PierreJeanjacquot: yes! It will make it clearer

PR #126, opened 2024-01-17
split monorepo pipelines - develop branch

PierreJeanjacquot:
- split pipelines by package
- rename promote targets to include the package name


Le-Caignec (approved):
LGTM 🚀

PR #129, opened 2024-01-19
Feature/manage renting options

Le-Caignec:
Modify current `setProtectedDataToRenting` function to add renting options in the params of the function


SeddikBellamine (approved):
LGTM!

PR #132, opened 2024-01-22
Keep all types in "types.ts" for now

cedric25:


PierreJeanjacquot (approved):
LGTM, types split to do later

PR #138, opened 2024-01-24
[SDK] New `addToCollection()` function

cedric25:
Still to do:

- [ ] Try to get smart-contract error if transaction fails ("revert reason")
👉 We'll validate data as much as we can *before* calling the smart-contract. This seems to be good practice, mostly to avoid using gas.
- [ ] Can we check that something is a function with `yup`?

Creation of a custom `ErrorWithData` error. We can assess later if this is interesting or not.

---

### Choice to be made:
 - Only one call from UI = `addToCollection()` function will transparently call `protectData()` and `createCollection()`
 - 👉 2-3 calls from UI = first `protectData()` then maybe createCollection() and then `addToCollection()`

### Choice to be made:
How to give proper feedback to caller?
 - Use an observable like for `protectDataObservable`?
 - 👉 Allow caller to pass a callback function (ex: `addStatus`, `statusCallback`, `onStatusUpdate`, naming to be discussed)
Chosen naming for now: `onStatusUpdate()`


packages/sdk/.eslintrc.json
```diff
@@ -15,8 +15,9 @@
     "prettier"
   ],
   "rules": {
-    "sonarjs/cognitive-complexity": "warn",
     "@typescript-eslint/no-explicit-any": "warn",
+    "@typescript-eslint/no-use-before-define": ["error", { "functions": false }],
```
cedric25, line 19: Avoid ESLint error when using a function defined further in the file. (That works well thanks to hoisting.)

packages/sdk/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "scripts": {
     "build": "rm -rf dist && tsc --project tsconfig.build.json",
+    "build:watch": "npm run build -- --watch",
```
cedric25, line 17: Hot-reload for our SDK: Useful when working on the usecase-demo

packages/sdk/src/dataProtector/IExecDataProtector.ts
```diff
@@ -130,7 +131,7 @@ class IExecDataProtector {
   }
 
   transferOwnership(args: TransferParams): Promise<TransferResponse> {
-    return transferOwnership({ iexec: this.iexec, ...args });
+    return transferOwnership({ ...args, iexec: this.iexec });
```
cedric25, line 134: Simply putting `...args` at first like for other methods.

packages/sdk/src/dataProtector/IExecDataProtector.ts
```diff
@@ -17,16 +17,19 @@ import { protectData } from './protectData.js';
...
 import {
   AddressOrENS,
+  AddToCollectionParams,
   CreateCollectionResponse,
   DataProtectorConfigOptions,
   FetchGrantedAccessParams,
   FetchProtectedDataParams,
   GrantAccessParams,
   GrantedAccess,
+  GrantedAccessResponse,
```
cedric25, line 32: Just some alphabetical reordering

Le-Caignec (changes requested):
keep addStatus ✅
move to more singular functions ♻️

packages/sdk/src/dataProtector/types.ts
```diff
@@ -352,6 +357,23 @@ export type CreateCollectionResponse = {
   collectionId: number;
 };
 
+export type FetchCreatorProtectedDataParams = {
```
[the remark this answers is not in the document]
cedric25: Something that was not supposed to be commited in this PR, reverted.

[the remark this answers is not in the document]
Le-Caignec: no worries

packages/sdk/src/dataProtector/sharing/addToCollection.ts
```diff
@@ -0,0 +1,234 @@
...
+  // const vAddStatus: string = addressSchema().label('owner').validateSync(owner);
+
+  let protectedDataAddress = addressSchema()
+    .label('protectedDataAddress')
+    .validateSync(existingProtectedDataAddress);
+
+  if (protectedDataAddress) {
+    // TODO: Check that this given protected data belongs to the user
+  } else {
+    const createProtectedDataResult = await createProtectedData({
```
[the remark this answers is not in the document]
cedric25: I've just added a `checkAndGetProtectedData()` that checks that the given protected data exists and belongs to the connected user.

packages/sdk/src/dataProtector/sharing/addToCollection.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  if (protectedData.owner.id !== userAddress) {
+    throw new ErrorWithData('This protected data is not owned by the user.', {
+      protectedDataAddress,
+      userAddress,
+    });
+  }
+
+  return protectedData;
```
Le-Caignec, line 121: i think we didn't need to return anything here

packages/sdk/src/dataProtector/sharing/addToCollection.ts
```diff
@@ -0,0 +1,166 @@
...
+    sharingContractAddress,
+  });
+
+  const ownerAddress: Address = await (
+    collectionContract.connect(signer) as Contract
+  )
+    .ownerOf(collectionId)
+    .catch(() => {
+      // Do nothing
+      // An error means that the collection does not exist
```
Le-Caignec, line 148: to do in next PR 🤷🏻‍♂️ : 
```
} catch (error: any) {
    // catch the revert custom error : error ERC721NonexistentToken(uint256 tokenId);
    error = collectionContract.interface.decodeErrorResult(
      "ERC721NonexistentToken",
      error.data.data
    )[0];
    console.log("decoded error", error);```

Le-Caignec (approved):
LGTM some comments 🚀
I think checker (checkCollection, checkAndGetProtectedData) will be useful for many functions. May be it should be set in an other file.

packages/sdk/src/dataProtector/sharing/addToCollection.ts
cedric25: Yes that seems strange but that makes it convenient for the `approveCollectionContract` function (called later) to check the collection's owner.
I'll change the code to make it a bit more explicit.

packages/sdk/src/dataProtector/sharing/addToCollection.ts
cedric25: I've added that but commented for now (`error.data` is still undefined...)

PR #140, opened 2024-01-25
Feature/refactore smart contract architecture

Le-Caignec:
Refactoring with the new architecture


packages/sharing-smart-contract/contracts/Store.sol
```diff
@@ -69,27 +70,18 @@ abstract contract Store {
...
-            protectedDataInSubscription[_collectionId][_protectedData] == false,
-            "ProtectedData is available in subscription"
-        );
-        _;
-    }
     /***************************************************************************
-     *                       Renting                                      *
+     *                       Renting                                           *
      ***************************************************************************/
+    event ProtectedDataAddedToRenting(
```
luisccastillo, line 76: For renting is the correct term, in my opinion we would use the word 'to' for example, when adding the protected data to a collection 

```
The choice between "for renting" and "to renting" depends on the context and the intended meaning. However, it's important to note that "to renting" is not a standard phrase in English. The more common and grammatically correct expression is "for renting."

Here's a brief explanation:

For Renting:

This phrase is commonly used when describing something that is available or suitable for rent.
Example: "The apartment is for renting," or "The house is for renting."
In this context, "for" indicates the purpose or availability of the property for rent.
To Renting:

This construction is not typical in English, and it may be considered incorrect.
If you mean to convey the idea of renting something, it's more appropriate to use "for renting."
Example: Instead of saying "The car is to renting," you should say "The car is for renting.
```"

Comment (luisccastillo) [reply packages/sharing-smart-contract/contracts/Store.sol:76]: in short, 
for   + verb
to   + noun

File: packages/sharing-smart-contract/contracts/Store.sol [hunk 2]
```diff
@@ -103,38 +95,18 @@ abstract contract Store {
...
      ***************************************************************************/
+    event ProtectedDataAddedForSale(uint256 _collectionId, address _protectedData, uint112 _price);
+    event ProtectedDataRemovedFromSale(uint256 _collectionId, address _protectedData);
+    event ProtectedDataSold(uint256 _collectionIdFrom, address _protectedData, address _to);
+
     // collectionId => (protectedDataAddress: address => sellingParams: SellingParams)
     mapping(uint256 => mapping(address => SellingParams)) public protectedDataForSale;
 
     struct SellingParams {
         bool forSale;
```
luisccastillo, line 109: isForSale

packages/sharing-smart-contract/contracts/Store.sol
luisccastillo, line 1: isForRent for the RentingParams struct (like we agreed in slack)

packages/sharing-smart-contract/contracts/Store.sol
Le-Caignec: Commit: c71e128d179da32fb376e3f1d4f02199f3928bac

packages/sharing-smart-contract/contracts/Store.sol
Le-Caignec: Commit: c71e128d179da32fb376e3f1d4f02199f3928bac

Le-Caignec: yeah my bad

PR #147, opened 2024-02-02
index renting events

luisccastillo:


packages/subgraph/package.json
```diff
@@ -21,6 +22,7 @@
     "uuid": "^9.0.0"
   },
   "devDependencies": {
-    "@types/uuid": "^9.0.1"
+    "@types/uuid": "^9.0.1",
```
Le-Caignec: useless. It never use

packages/subgraph/src/protectedDataSharing.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+export function handleNewRental(event: NewRentalEvent): void {
+  const rental = new Rental(
+    (
+      event.params._protectedData.toI64() +
+      event.params.renter.toI64() +
+      event.block.timestamp.toI64() +
+      event.logIndex.toI64()
+    ).toString()
```
Le-Caignec: it causes indexing error. Indeed, it's impossible to cast hex to I64

packages/subgraph/subgraph.template.yaml
```diff
@@ -50,54 +50,46 @@ dataSources:
...
+      address: "0xaf6456760A41513F7A89321627E3BfeF47CEc753"
       abi: ProtectedDataSharing
       #START_BLOCK_PLACEHOLDER#
     mapping:
       kind: ethereum/events
       apiVersion: 0.0.7
       language: wasm/assemblyscript
       entities:
+        - Collection
         - CollectionSubscription
```
Le-Caignec: Add missing entitties

Le-Caignec (approved):
Fix with Commit: df654681533ce40fc34db244d7c59656131986fc

packages/subgraph/package.json
Le-Caignec: Commit: df654681533ce40fc34db244d7c59656131986fc

packages/subgraph/subgraph.template.yaml
Le-Caignec: Commit: df654681533ce40fc34db244d7c59656131986fc

PR #151, opened 2024-02-07
[sdk][subgraph][deserializer] borsh serialization

PierreJeanjacquot:
# Use borsh for data serialization

## Motivation

The `protectData` methods first serialize all the leaves of the data object into a binary representation and then package them in a zip file.
The current implementation of the serialization comes with some limitations:
- custom serialization spec with no existing deserialization libraries for dapps
- no support for float numbers
- no support for bigint

## Proposed solution

To address these limitations we propose changing the serialization specification to use [borsh (Binary Object Representation Serializer for Hashing)](https://borsh.io/)
Borsh uses well-defined types and has implementations in several languages. Floats are supported up to f64 and integers up to i128.

| js type | serializes to |
|---|---|
|boolean|borsh "bool"|
|string|borsh "string"|
|number|borsh "f64"|
|bigint (inside the i128 boundaries)|borsh "i128"|

## Impacts

- introduce new schemas names ('bool', 'f64', 'i128')
- reuse existing 'string' schema with risk of collision (legacy deserialization could deserialize a borsh serialized string incorrectly)
- dapps will need to handle possibly two serialization specifications or choose to stick only to support legacy or borsh serialized protected data

Impacted components services

- [x] dataprotector SDK: change serialization
- [x] subgraph: handle new schemas
- [ ] existing dapps: update the deserialization algorithm or use [IExecDataProtectorDeserializer](#IExecDataProtectorDeserializer)
   - [ ] web3mail https://github.com/iExecBlockchainComputing/web3mail-sdk/pull/123
   - [ ] builders dapps?

# IExecDataProtectorDeserializer

## Motivation

To consume a protected data into a TEE dapp, the dapp developer must handle the protected data deserialization to get the plain data.

This can be proposed to the builders as a helper package to reduce the burden of boilerplate code on the dapp developer.

With the introduction of a new serialization specification, abstracting the deserialization for the dapp developer will be even more valuable.

## Proposed solution

IExecDataProtectorConsumer is a helper class for js/ts dapps that enables deserializing a value of the protected data with a known schema.

Example usage:

```ts
import { IExecDataProtectorDeserializer } from '@iexec/dataprotector-deserializer';

async function main() {
    const deserializer = new IExecDataProtectorDeserializer();
    const email = await deserializer.getValue('email', 'string');
    // ...
}
```


packages/dataprotector-consumer/prepare-legacy-dataprotector.js
```diff
@@ -0,0 +1,14 @@
...
+console.log(`removing "exports" from ${packageJsonPath}`);
+
+const { exports, ...packageJson } = JSON.parse(
+  fs
+    .readFileSync(packageJsonPath, {
+      encoding: 'utf-8',
+    })
+    .toString()
+);
+fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson));
```
cedric25, line 14: Not sure I understand what this file does :)

Maybe add some comments at top of the file?

packages/dataprotector-consumer/.nvmrc
```diff
@@ -0,0 +1 @@
+14
```
cedric25, line 1: Any reason why this is so low? Because of borsh?

PierreJeanjacquot: the nodejs TEE runtime provided by Scone is currently node 14, so dapps are very likely to use node 14 for now

packages/dataprotector-consumer/prepare-legacy-dataprotector.js
PierreJeanjacquot: Yes, it strips the `exports` key from the `package.json` to allow us to import an otherwise not exported internal module.
I'll add this in comment.

PierreJeanjacquot: [2001c59](https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/151/commits/2001c59db05d5cbd3efb9f3b60ed8277b5de9f8a)

packages/dataprotector-consumer/README.md
```diff
@@ -0,0 +1,38 @@
...
+## Get started
+
+In your nodejs iExec dapp
+
+```js
+const dataprotectorConsumer = new IExecDataProtectorConsumer();
+
+const value1 = await dataprotectorConsumer.getValue('path.to.value1', 'bool');
+const value2 = await dataprotectorConsumer.getValue('path.to.value2', 'string');
+```
```
SeddikBellamine, line 38: may be we need to add a real example of iExec dapp that uses the consumer

packages/dataprotector-consumer/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+**yarn:**
+
+```sh
+yarn add @iexec/dataprotector-consumer
+```
+
+## Get started
+
+In your nodejs iExec dapp
```
SeddikBellamine, line 31: what about python compatibility ?

SeddikBellamine (commented):
nice work, i left some comments

packages/dataprotector-consumer/README.md
PierreJeanjacquot: this is a js package but the logic can be replicated in python with [borsh-construct-py](https://github.com/near/borsh-construct-py)

we can add this to the documentation

```py
from borsh_construct import Bool, String, F64, U128

# bool schema
serializedBool = Bool.build(True)
print("serializedBool", serializedBool)
deserializedBool = Bool.parse(serializedBool)
print("deserializedBool", deserializedBool)

# string schema
serializedString = String.build("foo bar")
print("serializedString", serializedString)
deserializedString = String.parse(serializedString)
print("deserializedString", deserializedString)

# f64 schema
serializedF64 = F64.build(1.00000000001)
print("serializedF64", serializedF64)
deserializedF64 = F64.parse(serializedF64)
print("deserializedF64", deserializedF64)

# i128 schema
serializedU128 = U128.build(1234567890)
print("serializedU128", serializedU128)
deserializedU128 = U128.parse(serializedU128)
print("deserializedU128", deserializedU128)
```

packages/dataprotector-consumer/README.md
PierreJeanjacquot: here is the web3mail dapp update PR https://github.com/iExecBlockchainComputing/web3mail-sdk/pull/123

packages/dataprotector-consumer/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+   * | `"number"` (legacy schema) | `number` |
+   * | any other value | `Uint8Array` |
+   *
+   * Usage:
+   *
+   * ```js
+   * const value = await dataprotectorConsumer.getValue("path.to.string.value", "string");
+   * ```
+   */
+  async getValue<T extends string>(
```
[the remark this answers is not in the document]
PierreJeanjacquot: I tried to split but it introduced more variables as `this` is used a lot in the `getValue` method I wasn't convinced of the benefits.
Plus the file is not that big (179 lines including comments and MD doc) and the class has a single method.
My take is to leave it in a single file.

SeddikBellamine (approved):

PR #152, opened 2024-02-07
Feature/set protected data as rentable & remove protected data as rentable methods

SeddikBellamine:
Objective: Add setProtectedDataAsRentable & removeProtectedDataAsRentable functions

done: 

develop setProtectedDataAsRentable() in SDK

develop removeProtectedDataAsRentable() in SDK

add tests


packages/sdk/src/dataProtector/sharing/removeProtectedDataAsRentable.ts
```diff
@@ -0,0 +1,41 @@
...
+  iexec = throwIfMissing(),
+  collectionTokenId = throwIfMissing(),
+  protectedDataAddress = throwIfMissing(),
+  sharingContractAddress,
+}: IExecConsumer & {
+  sharingContractAddress: AddressOrENS;
+} & RemoveProtectedDataAsRentableParams): Promise<RemoveProtectedDataAsRentableResponse> => {
+  //TODO:Input validation
+  const { provider, signer } = await iexec.config.resolveContractsClient();
+
```
cedric25, line 22: In order to be more confident about the following smart-contract call, we could check a few things here:

 - checkProtectedData() →
. Check that the protected data exists
. Check that the protected data is owned by the Sharing smart-contract
. Check that the protected data is in a collection owned by the user

 - checkCollection() →
. Check that the collection exists
. Check that the collection is owned by the user

and throw an explicit error if necessary, which would make it clearer for the builder.
I've done something similar for `addToCollection()`, throwing an `ErrorWithData` for example.
Tell me what you think.

packages/sdk/tests/e2e/sharing/setProtectedDataAsRentable.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        const { success } = await dataProtector.setProtectedDataAsRentable({
+          protectedDataAddress: result.address,
+          collectionTokenId: collectionId,
+          durationInSeconds: duration,
+          priceInNRLC: price,
+        });
+        expect(success).toBe(true);
+      },
+      10 * MAX_EXPECTED_BLOCKTIME
```
Le-Caignec, line 46:
```suggestion
      8 * MAX_EXPECTED_BLOCKTIME + MAX_EXPECTED_WEB2_SERVICES_TIME
```
In protectData there is web2 services as SMS ....

Le-Caignec: 4 TX so 8 * MAX_EXPECTED_BLOCKTIME

packages/sdk/tests/e2e/sharing/removeProtectedDataAsRentable.test.ts
```diff
@@ -0,0 +1,53 @@
...
+          durationInSeconds: duration,
+          priceInNRLC: price,
+        });
+        const { success } = await dataProtector.removeProtectedDataAsRentable({
+          collectionTokenId: collectionId,
+          protectedDataAddress: result.address,
+        });
+        expect(success).toBe(true);
+      },
+      15 * MAX_EXPECTED_BLOCKTIME
```
Le-Caignec, line 50:
```suggestion
      10 * MAX_EXPECTED_BLOCKTIME + MAX_EXPECTED_WEB2_SERVICES_TIME
```

packages/sdk/src/dataProtector/types.ts
```diff
@@ -373,6 +385,15 @@ export type SetProtectedDataToSubscriptionResponse = {
 export type SetSubscriptionOptionsResponse = {
   success: boolean;
 };
+
+export type SetProtectedDataAsRentableResponse = {
+  success: boolean;
```
Le-Caignec, line 390: We should add Tx as we discuss. To do in shadow 😉

packages/sdk/src/dataProtector/sharing/setProtectedDataAsRentable.ts
```diff
@@ -0,0 +1,48 @@
...
+  sharingContractAddress: AddressOrENS;
+} & SetProtectedDataAsRentableParams): Promise<SetProtectedDataAsRentableResponse> => {
+  //TODO:Input validation
+  const { provider, signer } = await iexec.config.resolveContractsClient();
+
+  const sharingContract = new ethers.Contract(
+    sharingContractAddress,
+    sharingABI,
+    provider
+  );
```
Le-Caignec, line 29:
```suggestion
  const sharingContract = await getSharingContract();
```

packages/sdk/src/dataProtector/sharing/removeProtectedDataAsRentable.ts
```diff
@@ -0,0 +1,41 @@
...
+  const sharingContract = new ethers.Contract(
+    sharingContractAddress,
+    sharingABI,
+    provider
+  );
```
Le-Caignec, line 27:
```suggestion
  const sharingContract = await getSharingContract();
```

packages/sdk/src/dataProtector/sharing/removeProtectedDataAsRentable.ts
Le-Caignec: Yeah i agree. In fact, you should implement all the require + modifier that is implemented in this function on the SC to avoid TX that will revert and return the appropriate message

Le-Caignec (changes requested):
Some Changes

Le-Caignec:
can you rename the function to `setProtectedDataToRenting` & `removeProtectedDataFromRenting`

packages/sdk/src/dataProtector/sharing/removeProtectedDataAsRentable.ts
SeddikBellamine: i added 2 checks : isCollectionOwner & isProtectedDataInCollection

packages/sdk/src/dataProtector/sharing/setProtectedDataToRenting.ts
```diff
@@ -0,0 +1,65 @@
...
+  const userAddress = await iexec.wallet.getAddress();
+  if (
+    !(await isCollectionOwner({
+      graphQLClient,
+      collectionTokenId: collectionTokenId,
+      walletAddress: userAddress,
+    }))
+  ) {
+    throw new WorkflowError(
+      'Failed to Set Protected Data To Renting: user is not collection owner.'
```
cedric25, line 35: This can be a misleading error message, it could also be that the collection hasn't been found.

packages/sdk/tests/e2e/sharing/setProtectedDataAsRentable.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        );
+      },
+      8 * MAX_EXPECTED_BLOCKTIME + MAX_EXPECTED_WEB2_SERVICES_TIME
+    );
+    it(
+      'should fail with protected data is not in collection error',
+      async () => {
+        const protectedDataAddressMock = Wallet.createRandom().address;
+        //create collection
+        const { collectionTokenId } = await dataProtector.createCollection();
```
cedric25, line 98: After this call we need to wait a few seconds for the subgraph to have time to index the event.
```
await sleep(2_000);
```

? 🤷‍♂️

Or setting this value as a constant as we'll reuse it in some other tests.
```
export const SUBGRAPH_EVENT_INDEXING_WAIT_TIME = 80_000;
```

packages/sdk/src/dataProtector/sharing/setProtectedDataToRenting.ts
SeddikBellamine: for that we need to add another check before this check if the collection is found or not.

packages/sdk/src/dataProtector/sharing/setProtectedDataToRenting.ts
```diff
@@ -0,0 +1,78 @@
...
+    !(await isProtectedDataInCollection({
+      graphQLClient,
+      protectedDataAddress,
+      collectionTokenId: collectionTokenId,
+    }))
+  ) {
+    throw new WorkflowError(
+      'Failed to Set Protected Data To Renting: Protected Data is not in collection.'
+    );
+  }
```
Le-Caignec, line 62: we should also verify that the protectedData is not available for Sale

Le-Caignec: add TODO (waiting for subgraph deployment)

Le-Caignec (approved):
Good Job

PR #156, opened 2024-02-08
feature/getRenters-method

Le-Caignec:


packages/demo/usecase-demo/src/modules/profile/SubscriptionOptionsForm.tsx
```diff
@@ -23,7 +23,7 @@ export function SubscriptionOptionsForm({
     }
 
     const dataProtector = await getDataProtectorClient();
-    await dataProtector.setSubscriptionOptions({
+    await dataProtector.setSubscriptionParams({
```
cedric25, line 26: 👍

packages/sdk/src/dataProtector/IExecDataProtector.ts
```diff
@@ -50,13 +51,19 @@ import {
...
   Taskid,
   TransferParams,
   TransferResponse,
   Web3SignerProvider,
+  GetRentersParams,
+  Renters,
+  AddToCollectionResponse,
+  Creator,
+  GetSubscribersResponse,
+  SubscribeResponse,
```
cedric25, line 66: Can we keep an alphabetical order in this list of imported members?

packages/sdk/src/dataProtector/IExecDataProtector.ts
```diff
@@ -205,23 +209,24 @@ class IExecDataProtector {
...
-  subscribe = (args: SubscribeParams) =>
+  subscribe = (args: SubscribeParams): Promise<SubscribeResponse> =>
     subscribe({
       ...args,
-      sharingContractAddress: this.sharingContractAddress,
-      iexec: this.iexec,
     });
 
-  getSubscribers = (args: SubscribeParams) =>
+  getSubscribers = (args: SubscribeParams): Promise<GetSubscribersResponse> =>
```
cedric25, line 217: Should we also have a `GetSubscribersParams` type?

I would say let's have a "CollectionIdParam" with `{ collectionId: number }` that we could reuse where needed.

And also a "SuccessResponse" with `{ success: true }`

packages/sdk/src/dataProtector/sharing/addToCollection.ts
```diff
@@ -90,6 +89,7 @@ export const addToCollection = async ({
     title: 'Add protected data to your collection',
     isDone: true,
   });
+  return { transaction: tx };
```
cedric25, line 92:
```suggestion
  return {
    success: true,
    transaction: tx
  };
```

wdyt?

packages/sdk/src/dataProtector/IExecDataProtector.ts
```diff
@@ -21,10 +21,11 @@ import { addToCollection } from './sharing/addToCollection.js';
 import { createCollection } from './sharing/createCollection.js';
 import { getSubscribers } from './sharing/getSubscribers.js';
 import { setProtectedDataToSubscription } from './sharing/setProtectedDataToSubscription.js';
-import { setSubscriptionOptions } from './sharing/setSubscriptionOptions.js';
+import { setSubscriptionParams } from './sharing/setSubscriptionParams.js';
 import { saveForSharingContract } from './sharing/smartContract/getSharingContract.js';
 import { getCollectionsByOwner } from './sharing/subgraph/getCollectionsByOwner.js';
 import { getCreators } from './sharing/subgraph/getCreators.js';
+import { getRenters } from './sharing/subgraph/getRenters.js';
```
cedric25, line 28: Just to confirm, are we ok to expose methods directly coming from a subgraph query? (I would say yes)

packages/sdk/src/dataProtector/IExecDataProtector.ts
Le-Caignec: Yep. I thought it was made automatically, the role of the config prettier to do that

packages/sdk/src/dataProtector/sharing/subgraph/getRenters.ts
```diff
@@ -0,0 +1,56 @@
...
+  graphQLClient,
+  protectedDataAddress = throwIfMissing(),
+}: {
+  graphQLClient: GraphQLClient;
+} & GetRentersParams): Promise<Renters[]> {
+  try {
+    const getRentersForProtectedDataQuery = gql`
+    query MyQuery {
+      protectedData(id: "${protectedDataAddress}") {
+        rentals {
```
cedric25, line 19: The common use-case will probably be to get **current** renters. Should we add a `endDate >= current date` filter?

Or a `includePastRentals: boolean` in parameters?

packages/sdk/src/dataProtector/sharing/subgraph/getRenters.ts
```diff
@@ -0,0 +1,56 @@
...
+      rentalParams: {
+        durationInSeconds: rental.rentalParams.duration,
+        priceInNRLC: rental.rentalParams.price,
+      },
+    }));
+
+    return renters;
+  } catch (error) {
+    console.error('Error fetching renters:', error);
+    return []; // Return empty array or handle error as per your application's requirement
```
cedric25, line 54: Usually not recommended to "hide" errors like that.
Probably better to `throw error` here.
Or simply removing try / catch block? 🤷‍♂️

Le-Caignec: Commit: ca05ad1d014bd0656849a362be553ea8537fa4c1

packages/sdk/src/dataProtector/IExecDataProtector.ts
Le-Caignec: yeah i agree with that. I think it could done in an other PR to avoid many conflicts with other PR not yet merged

packages/sdk/src/dataProtector/IExecDataProtector.ts
Le-Caignec: Yeah  think

packages/sdk/src/dataProtector/sharing/subgraph/getRenters.ts
Le-Caignec: yeah good questions. I prefer to include `includePastRentals: boolean` in parameters

Le-Caignec: Commit: 102f00b72b2374e77dc32bccc48b41416fc41804

cedric25 (approved):

PR #161, opened 2024-02-12
index protected data consumed event

luisccastillo:
- Added indexing for the consumption of protected data
- Corrected a bug where the state of a protectedData would be still for sale (isForsale) after being sold


cedric25:
`subgraph.template.yaml` line 75, still a `DealId` mentioned, should we replace it by `Consumption`?

cedric25 (approved):

cedric25 (approved):

PR #166, opened 2024-02-13
[SDK] `setProtectedDataForSale()` + `removeProtectedDataForSale()`

cedric25:


packages/sdk/src/dataProtector/sharing/removeProtectedDataForSale.ts
```diff
@@ -0,0 +1,115 @@
...
+  iexec = throwIfMissing(),
+  graphQLClient = throwIfMissing(),
+  protectedDataAddress,
+}: IExecConsumer &
+  SubgraphConsumer &
+  RemoveProtectedDataForSaleParams): Promise<SuccessWithTransactionHash> => {
+  addressSchema()
+    .required()
+    .label('protectedDataAddress')
+    .validateSync(protectedDataAddress);
```
cedric25, line 25: Little change here compared to existing code: I'm not created a new variable like `vProtectedDataAddress`. It looks like it's not necessary and validation will work as expected.

packages/sdk/src/dataProtector/IExecDataProtector.ts
```diff
@@ -242,6 +245,25 @@ class IExecDataProtector {
   rentProtectedData = (
     args: RentProtectedDataParams
   ): Promise<SuccessWithTransactionHash> => rentProtectedData(args);
+
+  setProtectedDataForSale = (args: {
+    protectedDataAddress: AddressOrENS;
+    priceInNRLC: number;
+  }): Promise<SuccessWithTransactionHash> => {
```
cedric25, line 252: Note that we don't require a `collectionTokenId` param since we can get it from the protected data.

packages/sdk/src/dataProtector/sharing/subgraph/getProtectedDataById.ts
```diff
@@ -15,13 +14,32 @@ export async function getProtectedDataById({
...
         }
+        collection {
+          id
+          owner {
+            id
+          }
+        }
+        isRentable
+        isIncludedInSubscription
+        isForSale
```
cedric25, line 29: I decided to reuse this subgraph function and just to add the fields I needed. It's fine for current callers of this method, everything will still work, they will just get unnecessary fields but that's ok.

packages/sdk/src/dataProtector/sharing/subgraph/getProtectedDataById.ts
```diff
@@ -15,13 +14,32 @@ export async function getProtectedDataById({
...
+    protectedData: {
+      id: Address;
+      name: string;
+      owner: { id: Address };
+      collection: { id: Address; owner: { id: Address } };
+      isRentable: boolean;
+      isIncludedInSubscription: boolean;
+      isForSale: boolean;
+    };
+  }>(getProtectedDataQuery);
```
cedric25, line 43: No need to extract a type and to reference it in `graphQLTypes.js` as it was only used in this file.
However we could extract this as a type and simply put it a the top of this file. Maybe cleaner?

packages/sdk/tests/test-utils.ts
```diff
@@ -82,6 +82,14 @@ export const MAX_EXPECTED_MARKET_API_PURGE_TIME = 5_000;
 
 export const MAX_EXPECTED_WEB2_SERVICES_TIME = 80_000;
 
+export const SUBGRAPH_CALL_TIMEOUT = 2_000;
+export const SMART_CONTRACT_CALL_TIMEOUT = 10_000;
+export const WAIT_FOR_SUBGRAPH_INDEXING = 2_000;
```
cedric25, line 87: I created some constants, I found them useful for my tests.. 🤷‍♂️

packages/sdk/src/dataProtector/sharing/setProtectedDataForSale.ts
```diff
@@ -0,0 +1,138 @@
...
+  const tx = await sharingContract.setProtectedDataForSale(
+    protectedData.collection.id,
+    protectedData.id,
+    priceInNRLC
+  );
+  const txReceipt = await tx.wait();
+
+  return {
+    success: true,
+    txHash: txReceipt.hash,
```
Le-Caignec, line 55: in the other function i return a `Transaction. It's an object that come from ethers

Le-Caignec: may be we can just return the hash. I don't know
But in this it should be `tx.hash` not `txReceipt.hash`

Le-Caignec (changes requested):
Somme changes. 
In the tests, the time out for TX should be 2*NB_TX*MAX_EXPECTED_BLOCKTIME. It's like a convention

packages/sdk/src/dataProtector/IExecDataProtector.ts
Le-Caignec: yeah good, idea we should verify that we can do the same thing in other functions that was already written

packages/sdk/src/dataProtector/sharing/setProtectedDataForSale.ts
cedric25: I think I checked, this is what's done most of the time: take the hash from the txReceipt.

You propose to change to:

```
  const tx = await sharingContract.setProtectedDataForSale(
    protectedData.collection.id,
    protectedData.id,
    vPriceInNRLC
  );
await tx.wait();

  return {
    success: true,
    txHash: tx.hash,
  };
```

ie. we don't care about the txReceipt, we just need to await the transaction.
And it's better to answer with the tx hash rather than with the receipt hash? (I guess they're two different hashes, right?)

packages/sdk/tests/e2e/sharing/removeProtectedDataForSale.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      name: 'test removeProtectedDataForSale()',
+    });
+    protectedDataAddress = address;
+
+    await dataProtector.addToCollection({
+      collectionTokenId,
+      protectedDataAddress,
+    });
+    await waitForSubgraphIndexing();
+  }, 3 * MAX_EXPECTED_BLOCKTIME + MAX_EXPECTED_WEB2_SERVICES_TIME);
```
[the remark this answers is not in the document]
cedric25: Chosen timeout here needs to account for:

**createCollection()**
One "write" call to a smart-contract

**protectData()**
One push to IPFS
One "write" call to a smart-contract (createDatasetWithSchema())
One push to SMS

**addToCollection()**
One call to subgraph (checkAndGetProtectedData())
One "read" call to a smart-contract (checkCollection())
One "write" call to a smart-contract (approveCollectionContract())
One "write" call to a smart-contract (addProtectedDataToCollection())

How do you get `6 * MAX_EXPECTED_BLOCKTIME + MAX_EXPECTED_WEB2_SERVICES_TIME` with that list?

MAX_EXPECTED_WEB2_SERVICES_TIME for the SMS call, if I'm correct.
2 * MAX_EXPECTED_BLOCKTIME for each "write" call to a smart-contract? So it should be 8?
No need to take the following into account? IPFS push, subgraph call, "read" call to smart-contract.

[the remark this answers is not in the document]
Le-Caignec: Yeah sorry you are right it's just 4*MAX_EXPECTED_BLOCKTIME + MAX_EXPECTED_WEB2_SERVICES_TIME

packages/sdk/src/dataProtector/sharing/setProtectedDataForSale.ts
Le-Caignec: yeah it's perfect like this. This is what i done in my PR also. I refactored the other function like this

packages/sdk/tests/e2e/sharing/removeProtectedDataForSale.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      // --- GIVEN
+      const invalidProtectedDataAddress = '0x123...';
+
+      // --- WHEN / THEN
+      await expect(
+        dataProtector.removeProtectedDataForSale({
+          protectedDataAddress: invalidProtectedDataAddress,
+        })
+      ).rejects.toThrow(
+        new ValidationError(
```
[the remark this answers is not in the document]
cedric25: Here it's an input validation error, it should be instant, no async code is really called, hence no timeout override here (it's 10sec by default)

But indeed I'm not sure if we should adapt the timeout like this for each test, or just set everywhere the timeout related to the tested function?

Le-Caignec (approved):

PR #173, opened 2024-02-19
[PortectedDataSharingContract] Fix first review hadrien

Le-Caignec:


packages/sharing-smart-contract/contracts/ProtectedDataSharing.sol
```diff
@@ -20,56 +20,34 @@ pragma solidity ^0.8.23;
...
-    mapping(address => mapping(address => uint48)) public renters;
-    // protectedData => last rental end timestamp
-    mapping(address => uint48) public lastRentalExpiration;
-
-    // ---------------------Sale state----------------------------------
-    // collectionTokenId => (protectedDataAddress: address => sellingParams: SellingParams)
-    mapping(uint256 => mapping(address => SellingParams)) public protectedDataForSale;
+
+    // protectedDataAddress => collectionTokenId
+    mapping(address => uint256) collectionForData;
```
Amxx, line 44: why is that not in `ProtectedDataDetails` ?

PR #174, opened 2024-02-20
[SDK] `buyProtectData()`

cedric25:


packages/sdk/src/lib/dataProtectorSharing/buyProtectedData.ts
```diff
@@ -0,0 +1,181 @@
...
+    if (collection) {
+      tx = await sharingContract.buyProtectedDataForCollection(
+        protectedData.collection.id, // _collectionTokenIdFrom
+        protectedData.id,
+        collection.id, // _collectionTokenIdTo
+        vAppAddress || DEFAULT_PROTECTED_DATA_SHARING_APP,
+        {
+          // TODO: See how we can remove this
+          gasLimit: 900_000,
+        }
```
Le-Caignec, line 82: you missed the value field : 
```suggestion
        {
           value: // price of the sale
          // TODO: See how we can remove this
          gasLimit: 900_000,
        }
```

packages/sdk/src/lib/dataProtectorSharing/buyProtectedData.ts
```diff
@@ -0,0 +1,181 @@
...
+      );
+    } else {
+      tx = await sharingContract.buyProtectedData(
+        protectedData.collection.id, // _collectionTokenIdFrom
+        protectedData.id,
+        userAddress,
+        {
+          // TODO: See how we can remove this
+          gasLimit: 900_000,
+        }
```
Le-Caignec, line 92: idem

Le-Caignec (changes requested):
Good Job, just some comments 👍🏻

packages/sdk/src/lib/dataProtectorSharing/buyProtectedData.ts
cedric25: Good catch indeed, thanks

Le-Caignec (approved):

PR #177, opened 2024-02-26
Feature/local stack tests

PierreJeanjacquot:
# Setup a local stack for tests

- [x] local test stack
  - [x] healthchecks
  - [x] start/stop script for the stack
- [x] CI test stack
  - [x] healthchecks
- [x] use local stack in tests

## Stack:

- blockchain
  - `bellecour-fork`
- core stack
  - `market-api` & `market-watcher`
  - `sms`
  - `result-proxy`
- dataprotector components 
  - `graphnode` with Dataprotector subgraph
  - `ipfs`
- ...and services deps (mutualized when possible)

The stack is configured by the `prepare-test-env.js` script to get the current block of the bellecour node to fork from.

### Local:

Before running tests locally, you must configure and start the local test stack by running

```sh
npm run start-test-stack
```
You will get a local stack running in docker.

Run your tests as usual
```sh
npm run tests
```

When finished you can teardown the local test stack
```sh
npm run stop-test-stack
```

If you need to refresh the stack (basically if the bellecour-fork restart, it will lose the history and components will be desynced), run the `start-test-stack` once again, this will teardown any existing stack and recreate a new one based on the current blockchain state.

### CI:

Do the same things as local but in a way we can do it in drone.

### DIFF
--- a/packages/sdk/.drone.yml
+++ b/packages/sdk/.drone.yml
@@ -7,6 +7,183 @@ trigger:
   event: push
 
 steps:
+  ##### START LOCAL STACK #####
+  - name: prepare local-stack-env
+    image: bash
+    commands:
+      - mkdir local-stack-env
+      - echo -n "https://bellecour.iex.ec" > local-stack-env/BELLECOUR_FORK_URL
+      # TODO get last block to configure BELLECOUR_FORK_BLOCK
+      - echo -n "26814000" > local-stack-env/BELLECOUR_FORK_BLOCK
+
+  - name: ipfs
+    image: ipfs/go-ipfs:v0.9.1
+    detach: true
+    expose:
+      - 8080
+      - 5001
+
+  - name: market-mongo
+    image: mongo:6.0.3
+    detach: true
+    expose:
+      - 27017
+
+  - name: market-redis
+    image: redis:7.0.7-alpine
+    detach: true
+    commands:
+      - redis-server --appendonly yes
+    expose:
+      - 6379
+
+  - name: result-proxy-mongo
+    image: library/mongo:4.2
+    detach: true
+    command:
+      - mongod --bind_ip_all --port 13202
+    expose:
+      - 13202
+
+  - name: graphnode-postgres
+    image: postgres:12
+    detach: true
+    commands:
+      - postgres -cshared_preload_libraries=pg_stat_statements
+    expose:
+      - 5432
+    environment:
+      POSTGRES_USER: graphnode
+      POSTGRES_PASSWORD: password
+      POSTGRES_DB: graphnode-db
+
+  - name: bellecour-fork
+    detach: true
+    image: ghcr.io/foundry-rs/foundry:latest
+    pull: always
+    expose:
+      - 8545
+    commands:
+      - anvil --host 0.0.0.0 --port 8545 --block-time 1 --hardfork berlin --fork-url $(cat local-stack-env/BELLECOUR_FORK_URL) --fork-block-number $(cat local-stack-env/BELLECOUR_FORK_BLOCK) --chain-id 134 --gas-limit 6700000 --gas-price 0
+
+  - name: deps healthchecks
+    image: bash
+    commands:
+      # TODO bellecour-fork healthcheck
+      - echo "bellecour-fork ready"
+      # TODO ipfs healthcheck
+      - echo "ipfs ready"
+      # TODO market-mongo healthcheck
+      - echo "market-mongo ready"
+      # TODO market-redis healthcheck
+      - echo "market-redis ready"
+      # TODO result-proxy-mongo healthcheck
+      - echo "result-proxy-mongo ready"
+      # TODO graphnode-postgres healthcheck
+      - echo "graphnode-postgres ready"
+      - sleep 30
+
+  - name: sms
+    image: iexechub/iexec-sms:7.1.0
+    detach: true
+    environment:
+      TZ: Europe/Paris
+      IEXEC_SMS_BLOCKCHAIN_NODE_ADDRESS: http://bellecour-fork:8545
+      IEXEC_HUB_ADDRESS: '0x3eca1B216A7DF1C7689aEb259fFB83ADFB894E7f'
+      IEXEC_TEE_WORKER_PRE_COMPUTE_IMAGE: docker.io/iexechub/tee-worker-pre-compute:7.1.0-sconify-5.3.15-debug
+      IEXEC_TEE_WORKER_PRE_COMPUTE_FINGERPRINT: 9f0f782d6edc611baa23ca0978f555ee58ea70e092640c961e75c25e9e4b0f22
+      IEXEC_TEE_WORKER_PRE_COMPUTE_HEAP_SIZE_GB: 4
+      IEXEC_TEE_WORKER_POST_COMPUTE_IMAGE: docker.io/iexechub/tee-worker-post-compute:7.1.1-sconify-5.3.15-debug
+      IEXEC_TEE_WORKER_POST_COMPUTE_FINGERPRINT: face1376b97131e2dc75a556381d47a2e03bed9e1bc11e462471f99d1eefae50
+      IEXEC_TEE_WORKER_POST_COMPUTE_HEAP_SIZE_GB: 4
+      IEXEC_IGNORED_SGX_ADVISORIES: INTEL-SA-00161,INTEL-SA-00289,INTEL-SA-00334,INTEL-SA-00381,INTEL-SA-00389,INTEL-SA-00220,INTEL-SA-00270,INTEL-SA-00293,INTEL-SA-00320,INTEL-SA-00329,INTEL-SA-00477
+      IEXEC_SCONE_TOLERATED_INSECURE_OPTIONS: debug-mode,hyperthreading,outdated-tcb
+      IEXEC_SMS_DISPLAY_DEBUG_SESSION: 'true'
+      IEXEC_SCONE_CAS_HOST: foo
+      IEXEC_SMS_IMAGE_LAS_IMAGE: foo
+    expose:
+      - 13300
+
+  - name: market-watcher
+    image: iexechub/iexec-market-watcher:6.4
+    detach: true
+    environment:
+      CHAIN: BELLECOUR
+      ETH_WS_HOST: ws://bellecour-fork:8545
+      ETH_RPC_HOST: http://bellecour-fork:8545
+      MONGO_HOST: market-mongo
+      REDIS_HOST: market-redis
+    commands:
+      - export START_BLOCK=$(cat local-stack-env/BELLECOUR_FORK_BLOCK)
+      - npm start
+
+  - name: market-api
+    image: iexechub/iexec-market-api:6.4
+    detach: true
+    expose:
+      - 3000
+    environment:
+      CHAINS: BELLECOUR_FORK
+      BELLECOUR_FORK_ETH_RPC_HOST: http://bellecour-fork:8545
+      BELLECOUR_FORK_CHAIN_ID: 134
+      BELLECOUR_FORK_IS_NATIVE: 'true'
+      BELLECOUR_FORK_IEXEC_ADDRESS: '0x3eca1B216A7DF1C7689aEb259fFB83ADFB894E7f'
+      MONGO_HOST: market-mongo
+      REDIS_HOST: market-redis
+      RATE_LIMIT_MAX: 10000
+      RATE_LIMIT_PERIOD: 60000
+      MAX_OPEN_ORDERS_PER_WALLET: 1000
+
+  - name: result-proxy
+    image: iexechub/iexec-result-proxy:7.1.0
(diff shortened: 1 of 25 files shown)


packages/sdk/.drone.yml
```diff
@@ -7,6 +7,216 @@ trigger:
...
+      MONGO_HOST: result-proxy-mongo
+      MONGO_PORT: 13202
+      IEXEC_IPFS_HOST: ipfs
+    expose:
+      - 13200
+    commands:
+      # bellecour-fork healthcheck
+      # result-proxy-mongo healthcheck
+      # ipfs healthcheck
+      - sleep 30
```
PierreJeanjacquot, line 134: no access to `nc` or `curl` in this image, so the classic open port check is replaced by `sleep 30`

packages/sdk/.drone.yml
```diff
@@ -7,6 +7,216 @@ trigger:
...
+      IEXEC_IGNORED_SGX_ADVISORIES: INTEL-SA-00161,INTEL-SA-00289,INTEL-SA-00334,INTEL-SA-00381,INTEL-SA-00389,INTEL-SA-00220,INTEL-SA-00270,INTEL-SA-00293,INTEL-SA-00320,INTEL-SA-00329,INTEL-SA-00477
+      IEXEC_SCONE_TOLERATED_INSECURE_OPTIONS: debug-mode,hyperthreading,outdated-tcb
+      IEXEC_SMS_DISPLAY_DEBUG_SESSION: 'true'
+      IEXEC_SCONE_CAS_HOST: foo
+      IEXEC_SMS_IMAGE_LAS_IMAGE: foo
+    expose:
+      - 13300
+    commands:
+      # bellecour-fork healthcheck
+      - sleep 30
```
PierreJeanjacquot, line 114: no access to `nc` or `curl` in this image, so the classic open port check is replaced by `sleep 30`

packages/sdk/package.json
```diff
@@ -19,7 +19,9 @@
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "format": "prettier --write \"(src|tests)/**/*.ts\"",
-    "check-format": "prettier --check \"(src|tests)/**/*.ts|tests/**/*.ts\""
+    "check-format": "prettier --check \"(src|tests)/**/*.ts|tests/**/*.ts\"",
+    "stop-test-stack": "cd tests && docker compose down --remove-orphans",
+    "start-test-stack": "cd tests && npm run stop-test-stack && node prepare-test-env.js && docker compose up -d"
```
PierreJeanjacquot, line 24: start and stop scripts (require docker compose)

packages/sdk/src/dataProtector/processProtectedData.ts
```diff
@@ -24,14 +25,15 @@ export const processProtectedData = async ({
...
 }: IExecConsumer & ProcessProtectedDataParams): Promise<Taskid> => {
   try {
     const requester = await iexec.wallet.getAddress();
-    const vApp = addressOrEnsOrAnySchema()
+    const vApp = addressOrEnsSchema()
       .required()
       .label('authorizedApp')
       .validateSync(app);
-    const vProtectedData = addressOrEnsOrAnySchema()
+    const vProtectedData = addressOrEnsSchema()
```
PierreJeanjacquot, line 36: this was a bug I found during the development, the protectedData must be specified in processProtectedData (cannot be "any")

packages/sdk/src/dataProtector/processProtectedData.ts
```diff
@@ -24,14 +25,15 @@ export const processProtectedData = async ({
   args,
   inputFiles,
   secrets,
+  workerpool,
```
PierreJeanjacquot, line 28: added `workerpool` option to override the default value as it was not possible (this is required for local testing as we don't control the default pool)

packages/sdk/src/dataProtector/protectDataObservable.ts
```diff
@@ -140,7 +140,8 @@ export const protectDataObservable = ({
             vName,
             JSON.stringify(schema),
             multiaddrBytes,
-            checksum
+            checksum,
+            txOptions
```
PierreJeanjacquot, line 144: txOptions specifies the `gasPrice` we want to inherit it from the `iexec` SDK instance

packages/sdk/src/utils/validators.ts
```diff
@@ -17,7 +17,8 @@ const isZeroStringTest = (value: string) => value === '0';
 export const stringSchema = () =>
   string().strict().typeError('${path} should be a string');
 
-export const urlSchema = () => string().url('${path} should be a url');
+export const urlSchema = () =>
+  string().matches(/^http[s]?:\/\//, '${path} should be a url');
```
PierreJeanjacquot, line 21: changed the url validation as `yup`'s one does not support `localhost`
the new implementation is not bulletproof but it provides a cheap quick validation

packages/sdk/tests/e2e/processProtectedData.test.ts
```diff
@@ -1,29 +1,76 @@
...
+      });
+    await resourceProvider.order
+      .createWorkerpoolorder({
+        workerpool: workerpoolAddress,
+        category: 0,
+        tag: ['tee', 'scone'],
+      })
+      .then(resourceProvider.order.signWorkerpoolorder)
+      .then(resourceProvider.order.publishWorkerpoolorder);
+    app = appAddress;
```
PierreJeanjacquot, line 66: create resources on the local stack as the one used before required orders published by third parties on the marketplace (our marketplace instance is empty)

packages/sdk/tests/prepare-test-env.js
PierreJeanjacquot, line 1: fetch the latest block of the bellecour blockchain to base the local fork

packages/sdk/tests/test-utils.ts
```diff
@@ -1,8 +1,46 @@
...
+    ipfsGateway: process.env.DRONE
+      ? 'http://ipfs:8080'
+      : 'http://localhost:8080',
+    ipfsNode: process.env.DRONE ? 'http://ipfs:5001' : 'http://localhost:5001',
+    subgraphUrl: process.env.DRONE
+      ? 'http://graphnode:8000/subgraphs/name/DataProtector'
+      : 'http://localhost:8000/subgraphs/name/DataProtector',
+  };
+  return [ethProvider, options];
+};
```
PierreJeanjacquot, line 42: create the configuration to connect the local test stack

packages/subgraph/deployer.Dockerfile
PierreJeanjacquot, line 1: provide a base image for subgraph deployment in local stack, this allow us to cache dependencies installation across multiple reinstantiation of the local stack.

packages/subgraph/schema.graphql
PierreJeanjacquot, line 1: some v2 code was left, rolled back to what it should be on develop branch (plus `Subscription` is reserved word that prevents to compile)

SeddikBellamine (changes requested):
Great work !! , i left you some comments, I really like the idea to change the used stack to a local one and not use the real stack.

packages/sdk/.drone.yml
```diff
@@ -7,6 +7,216 @@ trigger:
...
 steps:
+  ##### START LOCAL STACK #####
+  - name: prepare local-stack-env
+    image: node
+    commands:
+      - node packages/sdk/tests/prepare-test-env.js
+
+  - name: bellecour-fork
+    detach: true
+    image: ghcr.io/foundry-rs/foundry:latest
```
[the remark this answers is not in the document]
PierreJeanjacquot: I initially started with a hardhat node but had some trouble when all the stack components were connected to it (node becoming unresponsive)
However, I haven't tested hardhat again after stabilizing with foundry anvil.
Another benefit I found after tinkering with foundry was the ability to pass all the fork parameters through cli options, freeing us of the additional burden of maintaining the hardhat docker image with bellecour fork configuration.

packages/sdk/.drone.yml
```diff
@@ -7,6 +7,216 @@ trigger:
...
+      ipfs: ipfs:5001
+      ethereum: bellecour:http://bellecour-fork:8545
+    commands:
+      # bellecour-fork healthcheck
+      - while ! nc -z bellecour-fork 8545 ; do sleep 1 ; done && echo "bellecour-fork ready"
+      # ipfs healthcheck
+      - while ! nc -z ipfs 8080 ; do sleep 1 ; done && echo "ipfs gateway ready"
+      - while ! nc -z ipfs 5001 ; do sleep 1 ; done && echo "ipfs upload ready"
+      # graphnode-postgres healthcheck
+      - while ! nc -z graphnode-postgres 5432 ; do sleep 1 ; done && echo "graphnode-postgres ready"
```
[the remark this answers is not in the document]
PierreJeanjacquot: I'm aware of `depends_on` and I would like to use it the way I did for the local docker-compose counterpart.
For what I tested drone does not support depends_on X healthy (only depends on X started) which is not enough in this case.
There is probably some room for optimization with depends_on, detached services, and blocking healthchecks steps, but using `depends_on` would convert the whole pipeline into a graph pipeline which is slightly more complex, I'll ask the team for approval :+1:

packages/sdk/.drone.yml
```diff
@@ -7,6 +7,216 @@ trigger:
...
+      - docker-entrypoint.sh postgres -cshared_preload_libraries=pg_stat_statements
+    expose:
+      - 5432
+    environment:
+      POSTGRES_USER: graphnode
+      POSTGRES_PASSWORD: password
+      POSTGRES_DB: graphnode-db
+
+  - name: graphnode
+    image: graphprotocol/graph-node:v0.27.0
```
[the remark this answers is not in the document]
PierreJeanjacquot: sure, I'll check what is used in production

packages/sdk/src/dataProtector/processProtectedData.ts
```diff
@@ -88,7 +95,7 @@ export const processProtectedData = async ({
       appmaxprice: vMaxPrice,
```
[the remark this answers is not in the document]
PierreJeanjacquot: I think you found a bug :bug: :+1:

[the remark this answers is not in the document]
PierreJeanjacquot: fixed 69a1eb2df073675ee245907fe9b03486c658b829

packages/sdk/src/utils/validators.ts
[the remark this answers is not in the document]
PierreJeanjacquot: We also need to support drone/docker service name here (ex: `http://graphnode:8000`)
The `/^http[s]?:\/\//` regex is not that big maybe not worth adding in a variable as it is not used elsewhere

packages/sdk/.drone.yml
```diff
@@ -7,6 +7,216 @@ trigger:
...
+      - while ! nc -z market-mongo 27017 ; do sleep 1 ; done && echo "market-mongo ready"
+      # market-redis healthcheck
+      - while ! nc -z market-redis 6379 ; do sleep 1 ; done && echo "market-redis ready"
+      # start
+      - export START_BLOCK=$(cat local-stack-env/BELLECOUR_FORK_BLOCK)
+      - cd /app
+      - npm start
+
+  - name: local subgraph deployment
+    image: node:18
```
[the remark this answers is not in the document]
PierreJeanjacquot: There is some unclear reason why `18.19` is used in other pipelines probably linked to a bug affecting `tsc`. Changing to 18.19 but only for the sake of having the same version. We need to challenge it.

changed 1aa2e3ad959ae4c8bac6f17e2cbf93cea84b587c

packages/sdk/.drone.yml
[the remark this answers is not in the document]
PierreJeanjacquot: @SeddikBellamine proposal for a graphpipeline https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/184

packages/sdk/.drone.yml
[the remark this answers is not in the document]
PierreJeanjacquot: checked with infra team, the deployed graphnodes uses `0.27.0`, nothing to change on our side IMO.

packages/sdk/.drone.yml
PierreJeanjacquot: changed in #184

packages/sdk/.drone.yml
PierreJeanjacquot: changed in #184

packages/sdk/.drone.yml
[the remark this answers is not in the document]
PierreJeanjacquot: changed in #184

SeddikBellamine (approved):
LGTM !

packages/sdk/.drone.yml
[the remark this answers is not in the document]
SeddikBellamine: got it

PR #187, opened 2024-03-06
Feature/local test stack

SeddikBellamine:
# Setup a local stack for tests

- [x] local test stack
  - [x] healthchecks
  - [x] start/stop script for the stack
- [x] CI test stack
  - [x] healthchecks
- [x] use local stack in tests

## Stack:

- blockchain
  - `bellecour-fork`
- core stack
  - `market-api` & `market-watcher`
  - `sms`
  - `result-proxy`
- dataprotector components 
  - `graphnode` with Dataprotector subgraph
  - `ipfs`
- ...and services deps (mutualized when possible)

The stack is configured by the `prepare-test-env.js` script to get the current block of the bellecour node to fork from.

### Local:

Before running tests locally, you must configure and start the local test stack by running

```sh
npm run start-test-stack
```
You will get a local stack running in docker.

Run your tests as usual
```sh
npm run tests
```

When finished you can teardown the local test stack
```sh
npm run stop-test-stack
```

If you need to refresh the stack (basically if the bellecour-fork restart, it will lose the history and components will be desynced), run the `start-test-stack` once again, this will teardown any existing stack and recreate a new one based on the current blockchain state.

### CI:

Do the same things as local but in a way we can do it in drone.


Le-Caignec (changes requested):
need to find solution

PR #190, opened 2024-03-18
[SDK] fetch dataset orderbook in strict mode

luisccastillo:


packages/sdk/src/dataProtector/fetchGrantedAccess.ts
```diff
@@ -37,6 +37,8 @@ export const fetchGrantedAccess = async ({
       {
         app: vAuthorizedApp,
         requester: vAuthorizedUser,
+        isAppStrict: true,
+        isRequesterStrict: true,
```
PierreJeanjacquot, line 41: we don't want to hardcode `true`

we want to add options to the `fetchGrantedAccess` methods and let the builders decide what behavior they want
- `isUserStrict` default false
- `isAppStrict` default false

PierreJeanjacquot (changes requested):

package.json
PierreJeanjacquot, line 1: top level `package-lock.json` should not exist

PierreJeanjacquot (approved):
LGTM
Maybe add a test with the new params set to true.

PR #196, opened 2024-03-21
migrate tests running on prod

PierreJeanjacquot:


packages/sdk/tests/e2e/dataProtectorCore/processProtectedData.test.ts
```diff
@@ -1,31 +1,61 @@
...
+    workerpoolAddress = workerpool;
+    await iexec.order
+      .createWorkerpoolorder({
+        workerpool: workerpoolAddress,
+        category: 0,
+        volume: 1000,
+        tag: ['tee', 'scone'],
+      })
+      .then(iexec.order.signWorkerpoolorder)
+      .then(iexec.order.publishWorkerpoolorder);
```
Le-Caignec, line 50: may be we can use a Promise.all here 🤷🏻‍♂️

Le-Caignec (approved):
one remarks

packages/sdk/tests/e2e/dataProtectorCore/processProtectedData.test.ts
PierreJeanjacquot: the same wallet is used in 2 tx, trying to parallelize will result in nonce collision risk, the second option is to use different wallets.

PR #204, opened 2024-03-27
Rename core types file + move DataProtectorConfigOptions to commonTypes

cedric25:
- Rename `dataProtectorTypes.ts` to `coreTypes.ts`, to be more consistent with the existing `sharingTypes.ts`

 - Move main `IExecDataProtector` constructor types (**DataProtectorConfigOptions**) to `commonTypes.ts` instead of `coreTypes.ts`.


Le-Caignec (approved):
LGTM Good 🚀

PR #208, opened 2024-03-28
`getRentals()`

cedric25:
We have a `getProtectedDataRentals()` that give us all rentals related to a protected data.

We also need a method to get all rentals related to a specific user.

I decided to transform the existing method and make it more versatile, ie. accept both a `protectedDataAddress` and a `userAddress` as per its input parameters.


packages/sdk/src/lib/dataProtectorSharing/getRentals.ts
```diff
@@ -0,0 +1,35 @@
...
+  const vIncludePastRentals = booleanSchema()
+    .label('includePastRentals')
+    .validateSync(includePastRentals);
+
+  return getRentalsQuery({
+    graphQLClient,
+    renterAddress: vRenterAddress,
+    protectedDataAddress: vProtectedDataAddress,
+    includePastRentals: vIncludePastRentals,
+  });
```
cedric25, line 34: No GraphQL response transformation here.
I'm trying to stay as close as possible to the GraphQL schema (Makes things easier regarding typing for example)

packages/sdk/src/lib/dataProtectorSharing/subgraph/getRentalsQuery.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const getRentalsQuery = async ({
+  graphQLClient = throwIfMissing(),
+  renterAddress,
+  protectedDataAddress,
+  includePastRentals = false,
+}: SubgraphConsumer & GetRentalsParams): Promise<GetRentalsGraphQLResponse> => {
+  const rentalsQuery = gql`
+    query ($where: Rental_filter) {
+      rentals(where: $where) {
```
cedric25, line 15: Simpler "where" here in query.
Conditional filters are applied in the "variables" down below. I found it cleaner than conditions in this template string (like in `getProtectedDataInCollectionsQuery.ts` where it's... not that clean)

packages/sdk/src/lib/dataProtectorSharing/getRentals.ts
```diff
@@ -0,0 +1,40 @@
...
+    .label('includePastRentals')
+    .validateSync(includePastRentals);
+
+  try {
+    return await getRentalsQuery({
+      graphQLClient,
+      renterAddress: vRenterAddress,
+      protectedDataAddress: vProtectedDataAddress,
+      includePastRentals: vIncludePastRentals,
+    });
```
cedric25, line 36: No transformation here, we directly return what the subgraph gave us.

Previously:
 - Renaming `endDate` to `endDateTimestamp`
 - Renaming `duration` to `durationInSeconds`
 - Renaming `price` to `priceInNRLC`

These "transformations" costs us, for example we need to have two different types, almost the same but not entirely.

And I feel like we could name these fields the way we want directly in the subgraph schema, so these "transformations" feel weird here.

packages/sdk/src/lib/dataProtectorSharing/subgraph/getRentalsQuery.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { SubgraphConsumer } from '../../types/internalTypes.js';
+
+export const getRentalsQuery = async ({
+  graphQLClient = throwIfMissing(),
+  renterAddress,
+  protectedDataAddress,
+  includePastRentals = false,
+}: SubgraphConsumer & GetRentalsParams): Promise<GetRentalsGraphQLResponse> => {
+  const rentalsQuery = gql`
+    query ($where: Rental_filter) {
```
Le-Caignec, line 14: where does Rental_filter type come from ?

packages/sdk/src/lib/dataProtectorSharing/getRentals.ts
Le-Caignec: yeah i agree. We should make a PR before the release to modify subgraph field naming so

Le-Caignec (approved):
LGTM

Le-Caignec:
It better to remove ProtectedData from name of functions. A big work (doc + sdk) but it should be do i think

packages/sdk/src/lib/dataProtectorSharing/subgraph/getRentalsQuery.ts
cedric25: We can see it in the schema docs for example:

Capture d’écran 2024-03-29 à 10 36 34

cedric25: We can see it in the graphiql schema docs for example:

Capture d’écran 2024-03-29 à 10 38 45

PR #214, opened 2024-04-08
Release/sdk 0.5.2

PierreJeanjacquot:
### Added

- Added `workerpool` option for `processProtectedData` to override the workerpool to use
- Added a dockerized local stack for testing

### Changed

- Run tests on a local stack forked from bellecour
- Support ENS names for `owner` option of `fetchProtectedData`
- Fixed a bug that allowed `"any"` to be passed as `newOwner` to `transferOwnership`
- Fixed a bug that allowed `"any"` to be passed as `protectedData` to `processProtectedData`
- Fixed a bug that allowed `"any"` to be passed as `app` to `processProtectedData`
- Fixed a bug that may cause the user to pay gas fees when creating a protected data
- Fixed a bug that may cause incorrect requestorder max prices in `processProtectedData`
- Fixed grantAccess to prevent address 0 to be used for authorizedApp
- Changed URL validation to be more permissive


packages/sdk/package.json
```diff
@@ -45,7 +47,7 @@
     "debug": "^4.3.4",
     "ethers": "^6.8.1",
     "graphql-request": "^6.0.0",
-    "iexec": "^8.5.0",
+    "iexec": "^8.6.0",
```
cedric25, line 50: Do we need to add a quick note on the changelog for this iexec-sdk bump?

packages/sdk/src/dataProtector/fetchGrantedAccess.ts
```diff
@@ -15,6 +16,8 @@ export const fetchGrantedAccess = async ({
   protectedData = 'any',
   authorizedApp = 'any',
   authorizedUser = 'any',
+  isAppStrict = false,
+  isUserStrict = false,
```
cedric25, line 20: We might need to mention these 2 new input parameters to the changelog?

packages/sdk/src/dataProtector/grantAccess.ts
```diff
@@ -52,6 +54,20 @@ export const grantAccess = async ({
...
+      throw new ValidationError('authorizedApp ENS name is not valid');
+    }
+    vAuthorizedApp = resolved.toLowerCase();
+  }
+
+  if (vAuthorizedApp === ZeroAddress) {
+    throw Error(
+      `Forbidden to use ${ZeroAddress} as authorizedApp, this would give access to any app`
+    );
+  }
```
cedric25, line 69: Should we mention this check in the changelog?

packages/sdk/src/dataProtector/fetchGrantedAccess.ts
PierreJeanjacquot: yes good catch

packages/sdk/package.json
PierreJeanjacquot: well... we didn't for the other upgrades of the iexec SDK, not sure what to do...

cedric25 (approved):

packages/sdk/package.json
cedric25: No problem then, I wasn't sure :)

PR #219, opened 2024-04-08
decrypt result on consume protected data

luisccastillo:
added a function in rsa.ts to format private keys to pem
added statuses for the protected data consumption
added workerpool as parameter for the consumeprotected data function 
result is decrypted with result-utils from the iexec sdk
result (unencrypted zip) is passed as an url for direct download


packages/sdk/src/lib/dataProtectorSharing/consumeProtectedData.ts
```diff
@@ -110,47 +116,90 @@ export const consumeProtectedData = async ({
...
+    const taskObservable = await iexec.task.obsTask(taskId, { dealid: dealId });
     vOnStatusUpdate({
-      title: 'UPLOAD_RESULT_TO_IPFS',
+      title: 'CONSUME_TASK_ACTIVE',
       isDone: true,
+      payload: {
+        taskId: taskId,
+      },
+    });
+    return await new Promise((resolve, reject) => {
```
Le-Caignec, line 143: Ca serait mieux de wrapper juste l'observable dans une promise qui resolve on complete, de l'await puis ensuite de reprendre le traitement hors de cette promise

Le-Caignec (approved):

packages/sdk/src/lib/dataProtectorSharing/consumeProtectedData.ts
Le-Caignec: fix Commit: 97e3055fb783dd0e8b73202dab1e9422fc79c4fd

PR #222, opened 2024-04-09
[Front] code review changes

ErwanDecoster:


packages/sdk/src/lib/types/sharingTypes.ts
```diff
@@ -106,6 +106,7 @@ export type Collection = {
 
 export type CollectionOwner = {
   id: Address;
+  hasActiveSubscription: boolean;
```
cedric25, line 109: Probably to be removed

packages/sdk/src/lib/types/graphQLTypes.ts
```diff
@@ -69,8 +69,15 @@ export type CollectionSubscription = {
 
 export type GetCollectionOwnersGraphQLResponse = {
   accounts: Array<{
+    hasActiveSubscription: boolean;
```
cedric25, line 72: Probably to be removed

packages/sdk/src/lib/dataProtectorSharing/getCollectionOwners.ts
```diff
@@ -4,25 +4,41 @@ import {
...
 
+    getCollectionOwnersQueryResponse.accounts.forEach((account) => {
+      account.hasActiveSubscription = account.collections.some(
+        (collection) =>
+          collection.subscriptions && collection.subscriptions.length > 0
+      );
+      account.collections.forEach(collection => {
+        delete collection.subscriptions;
+      })
+    })
```
cedric25, line 40: We'll try to make the same thing using `.map()` instead of `.forEach()`.

packages/demo/usecase-demo/src/components/ClickToExpand.tsx
```diff
@@ -0,0 +1,37 @@
...
+  className,
+}: {
+  title: string;
+  children: React.ReactNode;
+  className?: string;
+}) {
+  const [expanded, setExpanded] = useState(false);
+
+  return (
+    <div
```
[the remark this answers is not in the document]
ErwanDecoster: L'utilisation d'un bouton peut provoquer des conflits lors de l'envoi d'un formulaire s'il est placé à l'intérieur, comme c'est le cas dans CreateNewContent.tsx. Est-ce que je le déplace plus bas ?

packages/demo/usecase-demo/src/modules/DisclaimerModal.tsx
```diff
@@ -34,26 +33,31 @@ export function DisclaimerModal() {
...
 
-  if (!isStorageDiscaimerViewed) {
-    return (
-      <Dialog open={open} onOpenChange={setOpen}>
-        <DialogContent>
-          <DialogHeader>
+  return (
+    <Dialog open={open} onOpenChange={onModalOpenChange}>
+      <DialogContent className="pt-6">
+        <DialogDescription className="pb-6">
```
cedric25, line 50: Maybe just a `<div>` here as `DialogDescription` becomes a `p` tag, hence it should not contain divs.

packages/demo/usecase-demo/src/modules/DisclaimerModal.tsx
```diff
@@ -34,26 +33,31 @@ export function DisclaimerModal() {
...
               This dApp is a demo to learn the methods and see them in action.
               It is not a Content Creator application for creating and
               distributing Web3 content.
-            </DialogDescription>
-          </DialogHeader>
+            </p>
+          </div>
           <div className="grid gap-4">
-            <h4 className="text-xl font-bold">Try developper mode</h4>
+            <h4 className="text-xl font-bold">Try developer mode</h4>
```
cedric25, line 60: 👍

packages/demo/usecase-demo/src/components/useCarouselLogic.ts
```diff
@@ -0,0 +1,15 @@
...
+  });
+}
+
+export function OnScrollRight(carousel) {
+  carousel.current.scrollBy({
+    top: 0,
+    left: carousel.current.clientWidth,
+    behavior: 'smooth',
+  });
+}
```
cedric25, line 15: - functions should always start with lowercase letter: "OnScrollLeft" → "onScrollLeft"


- This doesn't look like a React hook, so the filename should not start with the keyword `use`. "useCarouselLogic.ts" → "carouselLogic.ts"

If it were a hook:
```
export function useCarouselLogic() {
  function OnScrollLeft(carousel) {
    // ...
  }

  function OnScrollRight(carousel) {
    // ...
  }

  return {
    OnScrollLeft,
    OnScrollRight,
  };
}

```


- This is not a component so I would move this file to our "utils" folder.


- Maybe name these functions just "scrollLeft()" and "scrollRight()"?


- Missing TypeScript type for `carousel` argument. Maybe: `React.RefObject<HTMLDivElement>`

packages/demo/usecase-demo/src/modules/ActiveRentals.tsx
```diff
@@ -33,8 +37,31 @@ export function ActiveRentals() {
...
+                </button>
+                <button
+                  className="group ml-1 p-1 transition-transform active:scale-[0.9]"
+                  onClick={() => OnScrollRight(rentedContent)}
+                >
+                  <div className="rounded-full bg-grey-700 p-2 transition-colors group-hover:bg-grey-500/40">
+                    <ArrowRight size="18" />
+                  </div>
+                </button>
+              </div>
```
cedric25, line 60: This code seems to be duplicated 3 times (`ActiveRentals.tsx`, `ActiveSubscriptions.tsx`, `ContentOfTheWeek.tsx`)
Maybe we can find a way to make it as a reusable component?

packages/demo/usecase-demo/src/components/ClickToExpand.tsx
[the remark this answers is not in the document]
cedric25: A `<button type="button">` should not interfere with a form as long as there is a `<button type="submit">` in the form, which is the case in "CreateNewContent.tsx". So I'd say it's fine?

> may cause conflicts

What did you have in mind exactly?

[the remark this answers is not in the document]
ErwanDecoster: ok perfect, thanks

packages/demo/usecase-demo/src/modules/ActiveRentals.tsx
ErwanDecoster: Nice suggestion, I'll do that in my next commit.

packages/demo/usecase-demo/src/utils/useCarouselLogic.ts
```diff
@@ -0,0 +1,22 @@
...
+      top: 0,
+      left: carousel.current.clientWidth,
+      behavior: 'smooth',
+    });
+  }
+  return {
+    scrollLeft,
+    scrollRight,
+  };
+}
```
cedric25, line 22: Only used by `CarouselScrollArrows`.
Not sure if it's worth it to have a separate hook here. I'd suggest to move these 2 functions directly to `CarouselScrollArrows.tsx` component.

PR #235, opened 2024-04-18
[DataProtectorSharing SC] Integrate fuzzing

Le-Caignec:


packages/sharing-smart-contract/test/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,28 @@
...
+import {HandlerSubscription} from "./handlers/HandlerSubscription.sol";
+import {HandlerRenting} from "./handlers/HandlerRenting.sol";
+
+contract Invariant is StdInvariant, Test {
+    function setUp() public {
+        vm.createSelectFork("https://bellecour.iex.ec");
+        HandlerSubscription hSubscription = new HandlerSubscription();
+        targetContract(address(hSubscription));
+
+        vm.createSelectFork("https://bellecour.iex.ec");
```
3docSec, line 16: a single `vm.createSelectFork` should be sufficient, no need to call it multiple times

packages/sharing-smart-contract/test/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,28 @@
...
+        HandlerRenting hRenting = new HandlerRenting();
+        targetContract(address(hRenting));
```
3docSec, line 18: are you sure this works as intended? I've seen a few times `targetContract` overwriting previous calls. So I suspect that with this code you'll only be testing the last one (`HandlerSale`)

packages/sharing-smart-contract/test/invariant/handlers/HandlerRenting.sol
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.24;
+
+import {IERC721} from "forge-std/interfaces/IERC721.sol";
+import {Test} from "forge-std/Test.sol";
+import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import {HandlerCollection} from "./HandlerCollection.sol";
+import {IRental} from "../../../contracts/interfaces/IRental.sol";
+
+contract HandlerRenting is Test, HandlerCollection {
```
3docSec, line 10: While I am in favor of organizing contracts with inheritance / composition, having handlers inherit `HandlerCollection` can be problematic: each handler will create their own data sets and will use it for its own purpose. For example, here, one data set can only be put for renting and rented, but this very same dataset will never be sold because the ghost tracking of `HandlerRenting is HandlerCollection` is different than `HandlerSale is HandlerCollection`.

It's like having a 12-cylinder sportscar that can simulate a gazillion scenarios, and you are driving it first gear :-)

PR #264, opened 2024-05-03
[SDK] Fix AddToCollection function

Le-Caignec:


packages/sdk/src/lib/dataProtectorSharing/addToCollection.ts
```diff
@@ -82,13 +82,15 @@ export const addToCollection = async ({
...
-  });
+  if (!approveTx) {
+    vOnStatusUpdate({
+      title: 'APPROVE_COLLECTION_CONTRACT',
+      isDone: true,
+      payload: {
+        approveTxHash: approveTx?.hash,
+      },
+    });
+  }
```
cedric25, line 93: If I read the code well, there will be no 'APPROVE_COLLECTION_CONTRACT' status sent if there is an approve transaction. Why?

---

To make the condition a bit clearer, we can extract it as a variable:

```
const hasContractAlreadyBeenApproved = !approveTx
```

Or I'd suggest to add a comment:
```
  // If no transaction, it probably means the contract has already been approved
  if (!approveTx) {
    // ...
  }
```

Le-Caignec: Fix commit : 57d976249a5ec253c2ebfa7300c24dea72568033

packages/sdk/tests/e2e/dataProtectorSharing/addToCollection.test.ts
```diff
@@ -50,6 +53,39 @@ describe('dataProtector.addToCollection()', () => {
         timeouts.createCollection +
         timeouts.addToCollection
     );
+
+    it(
+      'should work, if the protectedData has already been approved to the ProtectedDataSharing Contract',
```
PierreJeanjacquot, line 58:
```suggestion
      'should work, if the protectedData has already been approved to the DataProtectorSharing Contract',
```

cedric25 (approved):

PR #265, opened 2024-05-13
[SDK] fix callback function payload

Le-Caignec:


packages/sdk/src/lib/dataProtectorSharing/addToCollection.ts
```diff
@@ -83,17 +83,16 @@ export const addToCollection = async ({
...
-  }
+  vOnStatusUpdate({
+    title: 'APPROVE_COLLECTION_CONTRACT',
+    isDone: true,
+    payload: approveTx?.hash
+      ? { approveTxHash: approveTx.hash }
+      : {
+          message:
+            'Your ProtectedData has already been approved for the smart contract',
+        },
```
cedric25, line 94: I'm not sure we should write a message like this in English, the client will probably set its own message.
What we need is to give them something easy to check for, like a boolean.

```suggestion
      : { isAlreadyApproved: true },
```

or both:

```suggestion
      : {
          isAlreadyApproved: true,
          message:
            'Your ProtectedData has already been approved for the smart contract',
        },
```

I vote for this last option. The message adds more details, let's keep it.

cedric25 (approved):

PR #267, opened 2024-05-14
[SDK] Improve process protected data

Le-Caignec:
----------`ProcessProtectedData`-------------------------------
- [x] add `onStatusUpdate` in params of processProtectedData
- [x] adapt return types of `ProcessProtectedData` as `consumeProtectedData` function

----------`GetResultFromCompletedTask`-------------------------------
- [x] add validator for params
- [x] add a condition to decrypt or not the result if it's an encrypted one
- [x] add optional path in the function, if it's filled extract the file from the ZIP else return the Binary ZIP


packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -100,20 +141,72 @@ export const processProtectedData = async ({
...
 
     const requestorder = await iexec.order.signRequestorder(requestorderToSign);
 
-    const { dealid } = await iexec.order.matchOrders({
+    const { dealid, txHash } = await iexec.order.matchOrders({
       requestorder,
       ...underMaxPriceOrders,
     });
+    vOnStatusUpdate({
+      title: 'PROCESS_PROTECTED_DATA_REQUESTED',
```
cedric25, line 156: I'd like to keep a status with both `{ isDone: false }` then same status with `{ isDone: true }`

There is  no `{ isDone: false }` for this one.

Also I'd renamed it to: REQUEST_TO_PROCESS_PROTECTED_DATA
and that this event wraps 3 async calls:
- iexec.order.createRequestorder()
- iexec.order.signRequestorder()
- iexec.order.matchOrders()

wdyt?

packages/sdk/CHANGELOG.md
```diff
@@ -7,11 +7,13 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - `consumeProtectedData()`: Add two new status to ConsumeProtectedDataStatuses: "FETCH_WORKERPOOL_ORDERBOOK" and "PUSH_ENCRYPTION_KEY"
+- `processProtectedData()`: You can now pass it an `onStatusUpdate` callback to get similar "events" at each step of the process.
```
cedric25, line 10: I don't get the word "similar".
Oh because it refers to the previous sentence?

I would write:
```suggestion
- `processProtectedData()`: You can now pass an `onStatusUpdate` callback to get update events at each step of the process.
```

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -100,20 +141,72 @@ export const processProtectedData = async ({
...
+      isDone: true,
+      payload: {
+        txHash: txHash,
+      },
+    });
+
+    const taskId = await iexec.deal.computeTaskId(dealid, 0);
+    const taskObservable = await iexec.task.obsTask(taskId, { dealid: dealid });
+    vOnStatusUpdate({
+      title: 'CONSUME_TASK_ACTIVE',
```
cedric25, line 166: I've changed it for consumeProtectedData() to only have a "CONSUME_TASK" event with either `isDone: false` or `isDone: true`.

See PR https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/276/files

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -100,20 +141,72 @@ export const processProtectedData = async ({
...
+      taskObservable.subscribe({
+        next: () => {},
+        error: (e) => {
+          vOnStatusUpdate({
+            title: 'CONSUME_TASK_ERROR',
+            isDone: true,
+            payload: {
+              taskId: taskId,
+            },
+          });
```
cedric25, line 183: Unnecessary for me.
If an error is thrown, the builder deals with it in its catch block. Sending the error in an update event like this is redundant.

Also see PR https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/276/files

packages/sdk/src/lib/dataProtectorSharing/getResultFromCompletedTask.ts
```diff
@@ -2,72 +2,97 @@ import { decryptResult } from 'iexec/utils';
...
   GetResultFromCompletedTaskParams,
   GetResultFromCompletedTaskResponse,
+  GetResultFromCompletedTaskStatuses,
+  OnStatusUpdateFn,
 } from '../types/index.js';
 import { IExecConsumer } from '../types/internalTypes.js';
 
 export const getResultFromCompletedTask = async ({
   iexec = throwIfMissing(),
+  path,
```
cedric25, line 21: `taskId` seems more "important" to me, I'd put it before this new `path` arg.

packages/sdk/src/lib/types/coreTypes.ts
```diff
@@ -239,6 +239,18 @@ export type TransferResponse = {
...
 // ---------------------ProcessProtectedData Types------------------------------------
+export type ProcessProtectedDataStatuses =
+  | 'FETCH_PROTECTED_DATA_ORDERBOOK'
+  | 'FETCH_APP_ORDERBOOK'
+  | 'FETCH_WORKERPOOL_ORDERBOOK'
+  | 'PUSH_REQUESTER_SECRET'
+  | 'PROCESS_PROTECTED_DATA_REQUESTED'
+  | 'CONSUME_TASK_ACTIVE'
+  | 'CONSUME_TASK_ERROR'
+  | 'CONSUME_TASK_COMPLETED'
```
cedric25, line 250: We'll have only one here instead of these three: "CONSUME_TASK"

packages/sdk/src/lib/types/coreTypes.ts
```diff
@@ -277,4 +289,16 @@ export type ProcessProtectedDataParams = {
    * The workerpool to use for the application's execution. (default iExec production workerpool)
    */
   workerpool?: AddressOrENS | 'any';
+
+  /**
+   * Callback function ...
+   */
```
cedric25, line 295:
```suggestion
  /**
   * Callback function that will get called at each step of the process
   */
```

packages/sdk/tests/docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: '3.9'
-
```
cedric25, line 2: Why?

packages/sdk/src/utils/validators.ts
```diff
@@ -20,9 +20,18 @@ export const booleanSchema = () =>
 export const stringSchema = () =>
   string().strict().typeError('${path} should be a string');
 
+export const pathSchema = () =>
+  string().typeError('${path} should be a string');
```
cedric25, line 24: Why can't we reuse `stringSchema` instead of creating a new one?

packages/sdk/src/utils/validators.ts
```diff
@@ -20,9 +20,18 @@ export const booleanSchema = () =>
...
+
 export const urlSchema = () =>
   string().matches(/^http[s]?:\/\//, '${path} should be a url');
 
+export const taskIdSchema = () =>
+  string().matches(
+    /^0x[a-fA-F0-9]{64}$/,
+    '${path} must be a valid iExec task ID'
+  );
```
cedric25, line 33: Interesting new check! 👍

cedric25 (changes requested):

packages/sdk/src/lib/dataProtectorSharing/getResultFromCompletedTask.ts
```diff
@@ -2,72 +2,97 @@ import { decryptResult } from 'iexec/utils';
...
-  onStatusUpdate({
-    title: 'CONSUME_RESULT_DOWNLOAD',
-    isDone: false,
-  });
-  const taskResult = await iexec.task.fetchResults(taskId);
-  onStatusUpdate({
-    title: 'CONSUME_RESULT_DOWNLOAD',
-    isDone: true,
-  });
+  const vTaskId = taskIdSchema().label('taskId').validateSync(taskId);
```
cedric25, line 27:
```suggestion
  const vTaskId = taskIdSchema().required().label('taskId').validateSync(taskId);
```

packages/sdk/src/lib/dataProtectorSharing/getResultFromCompletedTask.ts
```diff
@@ -2,72 +2,97 @@ import { decryptResult } from 'iexec/utils';
...
-  if (!content) {
+async function extractFileFromZip(
+  resultBuffer: ArrayBuffer,
+  path: string
+): Promise<GetResultFromCompletedTaskResponse> {
+  const resultZip = await new JSZip().loadAsync(resultBuffer);
+  const file = resultZip.file(path);
+
+  if (!file) {
     throw new Error('No "content" file found in decrypted zip');
```
cedric25, line 93:
```suggestion
    throw new Error(`No "${path}" file found in decrypted zip`);
```

packages/sdk/tests/docker-compose.yml
Le-Caignec: It's depreciated now : 
`WARN[0000] /Users/robin/Documents/DataProtector/dataprotector-sdk/packages/sdk/tests/docker-compose.yml: `version` is obsolete`

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
Le-Caignec: commit: d06b52ac02242b9248b18303dc4ce9665dac4fe3

packages/sdk/CHANGELOG.md
Le-Caignec: Commit: f8b18749ea9a30e6541c90806035c94fb59da9e7

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
Le-Caignec: Commit: d87b9353854ec0d7169ccfe64c595f766ad00a73

packages/sdk/src/lib/dataProtectorSharing/getResultFromCompletedTask.ts
Le-Caignec: yeah i agree, moreover taskId is the only one that is not optional
Commit: e8a2427a9c2cba003785ce31f939a7daa9f719a2

packages/sdk/src/lib/types/coreTypes.ts
Le-Caignec: Commit: d87b9353854ec0d7169ccfe64c595f766ad00a73

packages/sdk/src/utils/validators.ts
Le-Caignec: Commit: f102d9059dad903eb40a70e612f1d0c87c56b0c9

cedric25 (approved):

PR #269, opened 2024-05-17
Re-allow boolean type

cedric25:
In `getProtectedData()`, we still need to be able to get older protected data that have been created with a type of `boolean`.


PierreJeanjacquot:
in #268 scope

PR #270, opened 2024-05-20
add voucher deployment on local test stack

luisccastillo:


packages/sdk/tests/test-voucher-utils.ts
```diff
@@ -0,0 +1,388 @@
...
+        params: [address, ethers.toBeHex(targetWeiBalance)],
+        id: 1,
+        jsonrpc: '2.0',
+      }),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+  } else {
+    await faucetSendWeiToReachTargetBalance(chain)(address, targetWeiBalance);
```
PierreJeanjacquot, line 114: will never be used, the whole `faucetSendWeiToReachTargetBalance` implementation can be removed

packages/sdk/tests/test-voucher-utils.ts
```diff
@@ -0,0 +1,388 @@
...
+          tryCount + 1
+        );
+      } else {
+        throw Error(`Failed to send RLC from faucet (tried ${tryCount} times)`);
+      }
+    }
+  };
+
+export const setNRlcBalance = (chain) => async (address, nRlcTargetBalance) => {
+  if (chain.isNative || chain.defaults?.isNative) {
```
PierreJeanjacquot, line 157: in this repo context, we always use the native asset as the protocol token, this is always true, the if statement can be removed as well as the whole `faucetSendNRlcToReachTargetBalance` implementation.

packages/sdk/tests/bellecour-fork/voucher-config.js
PierreJeanjacquot, line 1: convert to TS

PierreJeanjacquot (changes requested):
we can remove a lot of complexity that is useless in the context of this repo, see refactoring proposals in comments.

PR #271, opened 2024-05-20
[SDK] Feature/New reader constructor

Le-Caignec:
- New validator `isValidProvider` for "read / write" methods
- Better grouping of methods inside core and sharing interface files
- Validate `pemPublicKey` and `pemPrivateKey` for `consumeProtectedData()` method


packages/sdk/src/lib/IExecDataProtector.ts
```diff
@@ -11,7 +11,7 @@ class IExecDataProtector extends IExecDataProtectorModule {
   public sharing: IExecDataProtectorSharing;
 
   constructor(
-    ethProvider: Eip1193Provider | EnhancedWallet,
+    ethProvider?: Eip1193Provider | EnhancedWallet | string,
```
cedric25, line 14: Why do we add `string` here?

cedric25 (approved):

PR #277, opened 2024-05-23
Avoid useless rebuild of subgraph-deployer

PierreJeanjacquot:
Reduce the subgraph-deployer build context to avoid rebuilding the image when nothing related to the subgraph change

Benefits:
- faster first build due to reduced docker context
- reuse image from cache when nothing is changed in `environments` or `package/subgraph`


Le-Caignec (approved):
LGTM

PR #278, opened 2024-05-23
[SDK] Add new function setProtectedDataRentingParams

Le-Caignec:


cedric25 (changes requested):
Mention this new method to the changelog?

cedric25 (approved):

PR #283, opened 2024-06-05
add maxPrice parameter to consumeProtectedData

luisccastillo:


abbesBenayache (approved):
Add unit tests and end-to-end tests to test different scenarios

PR #306, opened 2024-07-05
[SDK] Remove any from function interfaces

Le-Caignec:
- Remove any from function interfaces
- Fixe tests

### DIFF
--- a/packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
+++ b/packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
@@ -7,7 +7,6 @@ import { WorkflowError } from '../../utils/errors.js';
 import { fetchOrdersUnderMaxPrice } from '../../utils/fetchOrdersUnderMaxPrice.js';
 import { pushRequesterSecret } from '../../utils/pushRequesterSecret.js';
 import {
-  addressOrEnsOrAnySchema,
   addressOrEnsSchema,
   positiveNumberSchema,
   secretsSchema,
@@ -33,7 +32,7 @@ export const processProtectedData = async ({
   args,
   inputFiles,
   secrets,
-  workerpool,
+  workerpool, // if the user want to specify all workerpool he should use Null_Address
   onStatusUpdate = () => {},
 }: IExecConsumer &
   ProcessProtectedDataParams): Promise<ProcessProtectedDataResponse> => {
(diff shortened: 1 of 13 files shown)


packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -33,7 +32,7 @@ export const processProtectedData = async ({
   args,
   inputFiles,
   secrets,
-  workerpool,
+  workerpool, // if the user want to specify all workerpool he should use Null_Address
```
PierreJeanjacquot, line 35: this will not work unless you also change the `fetchWorkerpoolOrderbook` call

- https://api.market.iex.ec/workerpoolorders?chainId=134&workerpool=0x0000000000000000000000000000000000000000 will always return an empty orderbook
- https://api.market.iex.ec/workerpoolorders?chainId=134&workerpool=any will return orders for any workerpools

```ts
    const workerpoolOrderbook = await iexec.orderbook.fetchWorkerpoolOrderbook({
      workerpool: vWorkerpool === ethers.ZeroAddress ? 'any' : vWorkerpool, // if address zero was chosen use any workerpool
      app: vApp,
      dataset: vProtectedData,
      minTag: SCONE_TAG,
      maxTag: SCONE_TAG,
});
```

PierreJeanjacquot (changes requested):

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
Le-Caignec: yeah it's true. 
Fix : Commit: a2bb109c04ff5025bc62bda5961e76210b6782e2
Add test : Commit: d92cf2975da9bc4081a402fba8d5fcb1e86a2b70

PS : I don't succeed to mock the getResultFromCompletedTask function import under the hood by processProtectedData. It's the same issue as consumeProtectedData.

packages/sdk/tests/unit/dataProtectorCore/processProtectedData.test.ts
```diff
@@ -200,4 +205,52 @@ describe('processProtectedData', () => {
...
+      expect(iexec.orderbook.fetchWorkerpoolOrderbook).toHaveBeenNthCalledWith(
+        1,
+        {
+          workerpool: 'any',
+          app: app,
+          dataset: protectedData.address.toLowerCase(),
+          minTag: SCONE_TAG,
+          maxTag: SCONE_TAG,
+        }
+      );
```
PierreJeanjacquot, line 252: try not to await the promise but the callback

PierreJeanjacquot (approved):

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
PierreJeanjacquot: will do in another PR

PierreJeanjacquot:
@Le-Caignec can you add this PR changes to the changelog?

PR #311, opened 2024-07-11
add user strict option on granted access

luisccastillo:


packages/sdk/tests/unit/dataProtectorCore/getGrantedAccess.test.ts
```diff
@@ -100,6 +102,7 @@ describe('getGrantedAccess', () => {
       {
         app: authorizedApp.toLowerCase(),
         requester: authorizedUser.toLowerCase(),
+        isRequesterStrict: false,
         page: 1,
         pageSize: 10,
       }
```
Le-Caignec, line 108: add one more test to check the true value

Le-Caignec (approved):
Need a release beta.5 for the doc

PR #314, opened 2024-07-15
Feature/process protected data usevoucher

luisccastillo:


packages/sdk/package-lock.json
PierreJeanjacquot, line 1: why so many changes here?

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -38,6 +39,7 @@ export const processProtectedData = async ({
   inputFiles,
   secrets,
   workerpool,
+  useVoucher,
```
PierreJeanjacquot, line 42: I would set the default value for `useVoucher` here and revert the change made to `dataProtectorCore/IExecDataProtectorCore.ts`

```suggestion
  useVoucher = false,
```

packages/sdk/src/lib/dataProtectorCore/IExecDataProtectorCore.ts
```diff
@@ -66,6 +66,7 @@ class IExecDataProtectorCore extends IExecDataProtectorModule {
     return processProtectedData({
       ...args,
       iexec: this.iexec,
+      useVoucher: args.useVoucher || false,
```
PierreJeanjacquot, line 69: the default value can be handled by `processProtectedData` implementation

```suggestion
```

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -154,10 +159,13 @@ export const processProtectedData = async ({
...
-    const { dealid, txHash } = await iexec.order.matchOrders({
-      requestorder,
-      ...underMaxPriceOrders,
-    });
+    const { dealid, txHash } = await iexec.order.matchOrders(
+      {
+        requestorder,
+        ...underMaxPriceOrders,
+      },
+      { preflightCheck: false, useVoucher: vUseVoucher }
```
PierreJeanjacquot, line 167: wy adding `preflightCheck: false`?
disabling preflightCheck can create bad deals that produce always falling tasks.

```suggestion
      { useVoucher: vUseVoucher }
```

PierreJeanjacquot (changes requested):
minor changes and topics to discuss

packages/sdk/package-lock.json
abbesBenayache: due to fixing iexec version and updating other packages after running npm install with Node.js 22

PierreJeanjacquot: ok, we should not do this in a feature PR

PierreJeanjacquot (approved):

PR #315, opened 2024-07-15
Return "multiaddr" field for protected data

cedric25:
In web3mail-usecase-demo, we say that IPFS link will be provided at the next version of the subgraph.
It has been like that for months. Either we remove this line, or we actually display the IPFS link. I believe we can, hence this PR.

Capture d’écran 2024-07-15 à 14 02 11

-----

TODO:
 - Return this new `multiaddr` also after `protectData()`


packages/sdk/tests/unit/utils/getMultiaddrAsString.test.ts
```diff
@@ -0,0 +1,51 @@
...
+        multiaddrAsHexString,
+      });
+
+      // --- THEN
+      expect(decodedMultiaddr).toEqual('');
+    });
+  });
+
+  describe('When giving a valid bytes value', () => {
+    it('should return a valid ipfs path', () => {
```
PierreJeanjacquot, line 35:
```suggestion
  describe('When giving a machine readable mutliaddr', () => {
    it('should return a human readable multiaddr', () => {
```

packages/sdk/src/utils/getMultiaddrAsString.ts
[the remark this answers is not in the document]
PierreJeanjacquot: I'm not sure we want to return an empty string in case of failure, this can trick the developer into thinking the multiaddr was decoded to `""`?
`undefined` may be better.

PierreJeanjacquot (changes requested):
- I would not return `/ipfs/` in human readable multiaddr since it is replaced by `/p2p/` in the multiaddr spec
- returning `multiaddr: ""` seems tricky, I would change the type to `multiaddr?: string`

PierreJeanjacquot (approved):

packages/sdk/src/lib/types/coreTypes.ts
```diff
@@ -76,6 +76,7 @@ export type ProtectedData = {
   owner: Address;
   schema: DataSchema;
   creationTimestamp: number;
+  multiaddr: string; // Ex: "/p2p/QmaiUykRQKPC2PDXvmiqdhDm553JybgLurUUiDYy78rMgY"
```
PierreJeanjacquot, line 79: may be undefined

```suggestion
  multiaddr?: string; // Ex: "/p2p/QmaiUykRQKPC2PDXvmiqdhDm553JybgLurUUiDYy78rMgY"
```

PR #320, opened 2024-07-18
[SDK] Allow grantAccess to authorize user restricted access when "any" user is authorized

Le-Caignec:


packages/sdk/tests/e2e/dataProtectorCore/grantAccess.test.ts
```diff
@@ -98,6 +98,25 @@ describe('dataProtectorCore.grantAccess()', () => {
...
+        ...input,
+        authorizedUser: ethers.ZeroAddress,
+        authorizedApp: sconeAppAddress,
+      });
+      // grantAccess to specific user
+      await dataProtectorCore.grantAccess({
+        ...input,
+        authorizedApp: sconeAppAddress,
+      });
+      expect(true).toBe(true);
```
abbesBenayache, line 115: you need to create a specific user and pass it as a parameter to later test if this user is in the grantedAccess list.

let test = expect(true).toBe(true); is always positive.

packages/sdk/tests/e2e/dataProtectorCore/grantAccess.test.ts
```diff
@@ -98,6 +98,25 @@ describe('dataProtectorCore.grantAccess()', () => {
...
+    },
+    2 * MAX_EXPECTED_WEB2_SERVICES_TIME
+  );
+
   it(
```
abbesBenayache, line 120: add tests for scenario 2 (pass isAppStrict as a parameter to the function `getGrantedAccess`)

packages/sdk/src/lib/dataProtectorCore/getGrantedAccess.ts
```diff
@@ -46,6 +46,7 @@ export const getGrantedAccess = async ({
         app: vAuthorizedApp || 'any',
         requester: vAuthorizedUser || 'any',
         isRequesterStrict: vIsUserStrict,
+        isAppStrict: true,
```
abbesBenayache, line 49: pass isAppStrict as a parameter to the function (use false as the default value) then validate:

const vIsAppStrict = booleanSchema()
  .label('isAppStrict')
  .validateSync(isAppStrict);

then call fetchDatasetOrderBook, passing isAppStrict = vIsAppStrict.

packages/sdk/tests/e2e/dataProtectorCore/grantAccess.test.ts
abbesBenayache: add test for displaying message : 

`An access has been already granted to the user: ${vAuthorizedUser} with the app: ${vAuthorizedApp}`

packages/sdk/src/lib/dataProtectorCore/getGrantedAccess.ts
PierreJeanjacquot: apprestrict address zero should not occur, the current implementation seems OK

packages/sdk/tests/e2e/dataProtectorCore/grantAccess.test.ts
Le-Caignec: `expect(true).toBe(true)` is always true, but it doesn't necessarily mean that the test will always succeed. Here, what we want to test is that the second call to the `grantAccess` function that does not revert. 

We use this quick win multiple times in [other tests as well](https://github.com/iExecBlockchainComputing/dataprotector-sdk/blob/dc1ae28835df9ec2462461cb3490e2544d705103/packages/sdk/tests/e2e/dataProtectorCore/revokeAllAccess.test.ts#L93).

However, it's true that we can also test the result of the function => Commit: 865291466d2189c08d03d9c3e6a08f737d56214a

packages/sdk/tests/e2e/dataProtectorCore/grantAccess.test.ts
Le-Caignec: `getGrantedAccess` should not have `isAppStrict` as new param => discuss this morning

Le-Caignec: Add new test to test error message thrown : 
Commit: 1306f3cb751c7f23e2103b1ef3d4625b5cd3db97

packages/sdk/tests/e2e/dataProtectorCore/grantAccess.test.ts
```diff
@@ -98,6 +98,54 @@ describe('dataProtectorCore.grantAccess()', () => {
...
+      });
+
+      expect(grantedAccess.requesterrestrict).toEqual(
+        input.authorizedUser.toLowerCase()
+      );
+    },
+    2 * MAX_EXPECTED_WEB2_SERVICES_TIME
+  );
+
+  it.only(
```
abbesBenayache, line 123: please remove the it.only from the test to ensure all tests are executed.


```suggestion
  it(
```

abbesBenayache (approved):
LGTM!

PR #322, opened 2024-07-19
fix script path for 'prepare-local-stack-env' step in CI config

abbesBenayache:


PierreJeanjacquot:
oops fix already pushed on voucher :grimacing:

PR #323, opened 2024-07-22
Add optional 'path' param to consumeProtectedData

cedric25:
The 'path' param indicates which file is to look for in the protected data zip.
No 'path' param = the entire zip will be returned as the result.

A 'path' param exists for `getResultFromCompletedTask()` but does not for `consumeProtectedData()`, which is kind of inconsistent.

On Content Creator usecase-demo, this inconsistency breaks the first render of the content once we have rented it.


packages/sdk/src/lib/dataProtectorSharing/consumeProtectedData.ts
```diff
@@ -39,9 +39,10 @@ export const consumeProtectedData = async ({
   iexec = throwIfMissing(),
   sharingContractAddress = throwIfMissing(),
   protectedData,
-  maxPrice = DEFAULT_MAX_PRICE,
   app,
+  path,
   workerpool,
+  maxPrice = DEFAULT_MAX_PRICE,
```
cedric25, line 45: Move this param later in the list of params, to be consistent with ordering in TypeScript type.

packages/sdk/src/lib/dataProtectorSharing/consumeProtectedData.ts
```diff
@@ -214,6 +215,7 @@ export const consumeProtectedData = async ({
     const { result } = await getResultFromCompletedTask({
       iexec,
       taskId,
+      path,
```
cedric25, line 218: Simply give this 'path' param to `getResultFromCompletedTask()` that knows what to do with it.

Le-Caignec (approved):
LGTM 🚀
Don't forget to add that to the doc 🙂

PR #329, opened 2024-07-25
Feature/iexec upgrade

PierreJeanjacquot:
# upgrade to iexec@8.10.0

- faster tx confirms in tests
- support for ethers AbstractProvider and AbstractSigner


Le-Caignec (approved):
LGTM

PR #331, opened 2024-07-29
Feature/view content without wallet

ErwanDecoster:


cedric25 (approved):
LGTM.

Maybe we could have extracted a new component with the OneContentCard modal, but not strictly required.

PR #332, opened 2024-07-31
More unit tests + dedicated step to run them

cedric25:
- Call the method directly. Do `getGrantedAccess()` instead of `dataProtectorCore.getGrantedAccess()`
Why?
This way it's easier to **inject** what we want through the iexec object. Indeed, the `iexec` object **is injected** into the method, so it can ease our tests.

- Create a new test file `getGrantedAccess.unit.ts`

- Move some tests out of `e2e/dataProtectorCore/getGrantedAccess.test.ts` that don't need the local stack to run (tests on input parameters, etc.)

- Be sure to have a good enough coverage with these unit tests.

- Review drone file: don't start expensive local stack until all "simple" steps are green.

- Also a new unit tests file for a sub-function: `formatGrantedAccess.unit.ts`


packages/sdk/.drone.yml
```diff
@@ -281,41 +337,7 @@ steps:
...
-      - cd packages/sdk
-      - npm run check-types
-    depends_on:
-      - install
-
-  - name: test
+  - name: test:e2e
     image: node:18.19
     commands:
       - cd packages/sdk
```
Le-Caignec, line 343: here we run again unit tests. May be we need also a script to make `npm run test:e2e`?

packages/sdk/package.json
```diff
@@ -64,11 +66,11 @@
...
   "devDependencies": {
-    "@jest/globals": "^29.5.0",
-    "@swc/core": "^1.3.96",
-    "@swc/jest": "^0.2.29",
+    "@jest/globals": "^29.7.0",
+    "@swc/core": "^1.7.3",
+    "@swc/jest": "^0.2.36",
     "@types/debug": "^4.1.7",
-    "@types/jest": "^29.5.11",
+    "@types/jest": "^29.5.12",
```
Le-Caignec, line 73: cool

Le-Caignec (approved):
LGTM
Good job 🙂

packages/sdk/tests/unit/dataProtectorCore/getGrantedAccess.unit.ts
```diff
@@ -0,0 +1,265 @@
+import { describe, expect, it, jest } from '@jest/globals';
+import { ApiCallError } from 'iexec/errors';
+import { ValidationError } from 'yup';
+import { WorkflowError } from '../../../src/index.js';
+import { getGrantedAccess } from '../../../src/lib/dataProtectorCore/getGrantedAccess.js';
+
+describe('getGrantedAccess', () => {
+  describe('Check validation for input parameters', () => {
+    describe('When given protected data is invalid', () => {
```
abbesBenayache, line 9: multiple `describe` blocks are used here, but many tests share the same configuration and context
we can use a single `describe` block for related tests to make tests suite easier to understand

packages/sdk/tests/unit/dataProtectorCore/getGrantedAccess.unit.ts
```diff
@@ -0,0 +1,265 @@
...
+          })
+          // --- THEN
+        ).rejects.toThrow(
+          new ValidationError(
+            'authorizedUser should be an ethereum address or a ENS name'
+          )
+        );
+      });
+    });
+
```
abbesBenayache, line 71: this test is repeated

packages/sdk/tests/unit/dataProtectorCore/getGrantedAccess.unit.ts
```diff
@@ -0,0 +1,265 @@
...
+          // --- WHEN
+          getGrantedAccess({
+            // @ts-expect-error No need for iexec here
+            iexec: {},
+            page: 0,
+            pageSize,
+          })
+          // --- THEN
+        ).rejects.toThrow(
+          new ValidationError('pageSize must be greater than or equal to 10')
```
abbesBenayache, line 87:
```suggestion
          new ValidationError(`pageSize must be greater than or equal to ${pageSize+1}`)
```

packages/sdk/tests/unit/utils/formatGrantedAccess.unit.ts
```diff
@@ -0,0 +1,36 @@
...
+      expect(formattedOrder).toEqual({
+        dataset: '0x35396912db97ff130411301ec722fc92ac37b00d', // <--
+        datasetprice: '0', // <--
+        volume: '10', // <--
+        tag: '0x0000000000000000000000000000000000000000000000000000000000000003',
+        apprestrict: '0x0000000000000000000000000000000000000000',
+        workerpoolrestrict: '0x0000000000000000000000000000000000000000',
+        requesterrestrict: '0x0000000000000000000000000000000000000000',
+        salt: '0x2a366726dc6321e78bba6697102f5953ceccfe6c0ddf9499dbb49c99bac1c16d',
+        sign: '0xb00707c4be504e6e07d20bd2e52babd72cbd26f064ec7648c6b684578232bee255a9c98aa2e9b18b4088602967d4f0641d52c0fbb3d5c00304a1f6d3c19eaf4f1c',
```
abbesBenayache, line 32: extract repeated mock data and configurations into fixtures

abbesBenayache (approved):

packages/sdk/.drone.yml
cedric25: Good point, I've added all possible unit + e2e test scripts to `package.json` and updated this drone config file accordingly.

packages/sdk/tests/unit/dataProtectorCore/getGrantedAccess.unit.ts
cedric25: > but many tests share the same configuration and context

What do you mean by "configuration and context"?
What piece of code could be extracted and moved to the top of the file?

packages/sdk/tests/unit/dataProtectorCore/getGrantedAccess.unit.ts
cedric25: The minimum value (10) is hard-coded in the actual validation code:
```
  const vPageSize = numberBetweenSchema(10, 1000)
    .label('pageSize')
    .validateSync(pageSize);
```

So I don't see why not using it directly in the test code.

What is the idea you want to highlight with `${pageSize+1}`?

PR #349, opened 2024-09-02
read registries addresses from poco contract instance

PierreJeanjacquot:
When using an instance of IExec PoCo different from the Bellecour one, the hardcoded addresses of registries were incorrect.
This PR replaces the hardcoded addresses with values read from the current instance of IExec PoCo.

NB: the used `getIExecContract()` method returns a contract with the [full ABI](https://github.com/iExecBlockchainComputing/iexec-sdk/blob/master/src/common/generated/%40iexec/poco/IexecInterfaceNative.js) (not limited to the one exposed in this repo)


cedric25 (approved):

Le-Caignec (commented):
LGTM 🚀

PR #354, opened 2024-09-11
Feature/custom update env

PierreJeanjacquot:
configure admin scripts with env variables

### DIFF
--- a/packages/sharing-smart-contract/README.md
+++ b/packages/sharing-smart-contract/README.md
@@ -31,24 +31,7 @@ To clean and compile the project:
 npm run compile
 ```
 
-### Verify
-
-To verify the contracts:
-
-```bash
-npm run verify
-```
-
-### Deploy (Production)
-
-To deploy the project on the production network - bellecour.
-⚠️ Be sure before deploying on bellecour
-
-```bash
-npm run script:prod
-```
-
-### Deploy (Test)
+### Deploy locally
 
 To deploy the project on the test network - localhost.
 You need first to start a local hardhat node which will be a fork of bellecour network :
@@ -60,7 +43,15 @@ npx hardhat node
 Open a new terminal and run :
 
 ```bash
-npm run script:test
+npm run deploy
+```
+
+### Upgrade locally
+
+upgrade the `DataprotectorSharing` contract implementation on a local bellecour fork.
+
+```sh
+npm run upgrade-local
 ```
 
 ### Run Tests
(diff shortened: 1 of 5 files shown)


packages/sharing-smart-contract/scripts/upgrade.js
```diff
@@ -1,16 +1,24 @@
...
   const { ENV } = process.env;
-  console.log(`Using ENV: ${ENV}`);
+  const loadFromEnv = getLoadFromEnv(ENV);
 
-  const { dataprotectorSharingContractAddress, addOnlyAppWhitelistRegistryContractAddress } = getEnvironment(ENV);
+  const {
+    DATAPROTECTOR_SHARING_ADDRESS = loadFromEnv('dataprotectorSharingContractAddress'),
+    ADD_ONLY_APP_WHITELIST_REGISTRY_ADDRESS = loadFromEnv('addOnlyAppWhitelistRegistryContractAddress'),
+    POCO_ADDRESS = defaultPocoAddress,
+    DATASET_REGISTRY_ADDRESS = defaultDatasetRegistryAddress,
```
Le-Caignec, line 20: why not load POCO_ADDRESS & DATASET_REGISTRY_ADDRESS from env also ? As those addresses could be different for bubble ?

packages/sharing-smart-contract/README.md
```diff
@@ -60,25 +66,73 @@ npx hardhat node
...
+upgrade the `DataprotectorSharing` contract implementation.
+
+```sh
+npm run upgrade -- --network <network>
+```
+
+Environment params:
+
+- `WALLET_PRIVATE_KEY`: DataprotectorSharing admin private key (bellecour network only)
+- `ENV`: default environment to use (`prod` or `staging`)
```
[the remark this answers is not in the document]
PierreJeanjacquot: not bubble, bubble is neither persistent nor public environment and is not meant to be added to the known environment

packages/sharing-smart-contract/scripts/upgrade.js
PierreJeanjacquot: Same, bubble will not be added to known environments as it constantly changes over redeployments and can be moved or shut down anytime, bubble should be treated as a custom ephemeral environment.
PoCo and DatasetRegistry address are not part of the environment package, they could be added but this is not needed for now because every environment is based on bellecour with a single instance of core contracts.
NB: this could change if staging moves on a dedicated chain or a dedicated instance of PoCo, but this is not something I have hands-on.

PierreJeanjacquot:
good questions, answered :)

PR #359, opened 2024-09-24
move try to return ValidationError

ErwanDecoster:


cedric25:
I think we can also update the CHANGELOG in this PR, adding that now `processProtectedData` and `getCollectionsByOwner` returns a ValidationError and not a WorkflowError anymore in case of a bad input parameter.

cedric25 (approved):

PR #364, opened 2024-09-30
Unit tests for transferOwnership()

cedric25:


packages/sdk/src/lib/dataProtectorCore/transferOwnership.ts
```diff
@@ -5,8 +5,8 @@ import { IExecConsumer } from '../types/internalTypes.js';
 
 export const transferOwnership = async ({
   iexec = throwIfMissing(),
-  protectedData = throwIfMissing(),
-  newOwner = throwIfMissing(),
+  protectedData,
+  newOwner,
```
cedric25, line 9: If not given, better to get:
```
protectedData should be an ethereum address or a ENS name
```

instead of:
```
Missing parameter
```

= Leverage `yup` validation library and have explicit error messages.

ErwanDecoster (approved):

PR #367, opened 2024-10-03
Unit tests for processProtectedData()

cedric25:


packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -33,10 +33,12 @@ import { IExecConsumer } from '../types/internalTypes.js';
 import { getWhitelistContract } from './smartContract/getWhitelistContract.js';
 import { isAddressInWhitelist } from './smartContract/whitelistContract.read.js';
 
+export type ProcessProtectedData = typeof processProtectedData;
```
cedric25, line 36: Type used for unit tests where method is dynamically imported with:

```
    testedModule = await import(
      '../../../../src/lib/dataProtectorCore/processProtectedData.js'
    );
    processProtectedData = testedModule.processProtectedData;
```

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -33,10 +33,12 @@ import { IExecConsumer } from '../types/internalTypes.js';
...
+
 export const processProtectedData = async ({
   iexec = throwIfMissing(),
-  protectedData = throwIfMissing(),
-  app = throwIfMissing(),
+  protectedData,
+  app,
```
cedric25, line 41: Let `yup` give us a proper "required" error instead of a very generic "Missing parameter".

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -46,14 +48,14 @@ export const processProtectedData = async ({
...
-    .label('authorizedApp')
-    .validateSync(app);
   const vProtectedData = addressOrEnsSchema()
     .required()
     .label('protectedData')
     .validateSync(protectedData);
+  const vApp = addressOrEnsSchema()
+    .required()
+    .label('authorizedApp')
+    .validateSync(app);
```
cedric25, line 58: Just move this. Have the same order than in the list of parameters.

packages/sdk/src/utils/validators.ts
```diff
@@ -97,7 +97,8 @@ export const grantedAccessSchema = () =>
     .noUnknown()
     .default(undefined);
 
-export const urlArraySchema = () => array().of(urlSchema());
+export const urlArraySchema = () =>
+  array().of(string().matches(/^http[s]?:\/\//, '${value} should be a url'));
```
cedric25, line 101: Specific array validator and use `${value}` here instead of `${path}`.
Otherwise we get something like "[0] is not a string" which is of limited help.

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -96,7 +98,7 @@ export const processProtectedData = async ({
 
         if (!isRequesterInWhitelist) {
           throw new Error(
-            `As a user, you are not in the whitelist. So you can't access to the protectedData in order process it`
+            "As a user, you are not in the whitelist. You can't access the protectedData so you can't process it."
```
cedric25, line 101: Fix some English here

packages/sdk/tests/e2e/dataProtectorCore/getGrantedAccess.test.ts
```diff
@@ -1,10 +1,10 @@
 import { describe, it, expect } from '@jest/globals';
 import { HDNodeWallet, Wallet } from 'ethers';
 import { IExecDataProtectorCore } from '../../../src/index.js';
+import { deployRandomApp } from '../../test-utils.e2e.js';
```
cedric25, line 4: This `deployRandomApp()` util method has moved to another file.

packages/sdk/tests/test-utils.e2e.ts
```diff
@@ -0,0 +1,34 @@
+import { Wallet } from 'ethers';
+import { IExecAppModule, TeeFramework } from 'iexec';
+import { getWeb3Provider, Web3SignerProvider } from '../src/index.js';
```
cedric25, line 3: This import was problematic for unit tests. What it does is that it loads the entire SDK so we couldn't mock some modules *before* loading the SDK.

That's why I moved this to a `test-utils.e2e.ts` file that is only used in e2e tests.

packages/sdk/tests/test-utils.ts
```diff
@@ -170,314 +138,25 @@ export const timeouts = {
   tx: 2 * MAX_EXPECTED_BLOCKTIME,
 };
 
-export function getOrderObject({
```
cedric25, line 173: Method renamed to `getDatasetOrderObject()` and moved to `utils/datasetOrders.ts`

packages/sdk/tests/unit/dataProtectorCore/grantAccess.test.ts
```diff
@@ -60,15 +63,15 @@ describe('dataProtectorCore.grantAccess()', () => {
...
+        const missingAuthorizedAppAddress = undefined;
 
         await expect(
           // --- WHEN
           grantAccess({
             // @ts-expect-error No need for iexec here
             iexec: {},
-            protectedData: '0xbb673ac41acfbee381fe2e784d14c53b1cdc5946',
-            authorizedApp: invalidAuthorizedAppAddress,
+            protectedData: getRandomAddress(),
```
cedric25, line 73: Better to just generate a random address.

packages/sdk/tests/unit/dataProtectorCore/processProtectedData/fetchOrdersUnderMaxPrice.test.ts
```diff
@@ -1,39 +1,34 @@
...
     ethProvider: getWeb3Provider(wallet.privateKey),
   });
 
-  const mockFetchWorkerpoolOrderbook: any = jest
-    .fn()
-    .mockImplementationOnce(() => {
-      return Promise.resolve(MOCK_WORKERPOOL_ORDER);
-    });
-  iexec.orderbook.fetchWorkerpoolOrderbook = mockFetchWorkerpoolOrderbook;
-
```
cedric25, line 25: Moved below line 27

packages/sdk/tests/utils/mockAllForProcessProtectedData.ts
```diff
@@ -0,0 +1,69 @@
+import { jest } from '@jest/globals';
```
cedric25, line 1: Complex file that will mock all necessary `iexec-sdk` methods for a `processProtectedData()` call.

packages/sdk/tests/test-utils.ts
```diff
@@ -2,10 +2,10 @@ import { jest } from '@jest/globals';
 import { Wallet, JsonRpcProvider, ethers } from 'ethers';
 import { IExecAppModule, IExecConfig, TeeFramework, utils } from 'iexec';
 import {
-  DataProtectorConfigOptions,
-  Web3SignerProvider,
-  getWeb3Provider,
+  type DataProtectorConfigOptions,
+  type Web3SignerProvider,
 } from '../src/index.js';
+import { getWeb3Provider } from '../src/utils/getWeb3Provider.js';
```
cedric25, line 8: Very important to **not** import this function from root `index.ts`.
Otherwise importing something from this `test-utils.ts` file would mean to import the whole SDK, which would break unit test mocks.

ErwanDecoster (approved):

PR #377, opened 2024-10-15
[Sharing-Smart-Contract] Stopt setting resultProxy

Le-Caignec:


Le-Caignec:
Units tests of the SDK are broken. This is why the CI is red. 
Not in the scope of this PR

PR #381, opened 2024-10-24
fix:  remove useless await

LeJamon:
Removed unnecessary await from generateSecureUniqueId() as the function is synchronous.


cedric25 (approved):
Indeed we can remove this `await` here, thanks for taking the time to open a Pull Request 🙌

PR #386, opened 2024-12-02
protectData add schema and zipFile in onStatusUpdate payload

PierreJeanjacquot:
expose `schema` and `zipFile` as soon as they are available through `onStatusUpdate`


PierreJeanjacquot:
currently not needed

PR #388, opened 2025-01-23
Feature/fix deserializer no protected data error

PierreJeanjacquot:
Fixed cryptic TypeError when `@iexec/dataprotector-deserializer` was used in a context without protected data, `getValue()` now rejects with Error "Missing protected data" in such a case.

Example with iapp hello-world template:

Before:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

After:
```
Error: Missing protected data
```


abbesBenayache (approved):
LGTM!

PR #389, opened 2025-01-27
Release/dataprotector deserializer v0.1.1

PierreJeanjacquot:
### Changed

- fixed cryptic TypeError when `@iexec/dataprotector-deserializer` was used in a context without protected data, `getValue()` now rejects with Error "Missing protected data" in such a case.


abbesBenayache (approved):
LGTM!

PR #390, opened 2025-01-27
Feature/allow debug

PierreJeanjacquot:
# add `allowDebug` option for `protectData`

When `protectData({ data, allowDebug: true })` is called, the protected data secret is pushed in both prod and debug SMS
`allowDebug` default value is `false`, and the default behavior does not change.

NB:
- `allowDebug: true` triggers an additional message signature request to push the secret on the debug SMS.
- `allowDebug` param description includes warnings about the risks of allowing TEE debug apps.

## misc

- debug SMS is configurable via `iexecOptions.smsDebugURL` a semi-hidden option (not documented) in the `DataprotectorModule` constructor
- debug SMS added to test stack
- protectData e2e test with png image fixed


packages/sdk/src/lib/types/internalTypes.ts
```diff
@@ -6,6 +6,13 @@ export type IExecConsumer = {
   iexec: IExec;
 };
 
+export type IExecDebugConsumer = {
```
Le-Caignec, line 9:
```suggestion
export type IExecDebugClient = {
```
Ca me parle plus 🤷🏻‍♂️

packages/sdk/src/lib/dataProtectorCore/protectData.ts
```diff
@@ -244,6 +248,35 @@ export const protectData = async ({
...
+        },
+      });
+      await iexecDebug.dataset
+        .pushDatasetSecret(protectedDataAddress, encryptionKey, {
+          teeFramework: 'scone',
+        })
+        .catch((e: Error) => {
+          handleIfProtocolError(e);
+          throw new WorkflowError({
+            message: 'Failed to push protected data encryption key',
```
Le-Caignec, line 267:
```suggestion
            message: 'Failed to push protected data encryption key into the debug SMS',
```

Le-Caignec (approved):
Some comments

packages/sdk/src/lib/dataProtectorCore/protectData.ts
PierreJeanjacquot: the error for prod SMS does not give this precision, I would not change the message for debug SMS alone, leaving it as is for now

PR #422, opened 2025-03-31
feature/fix-processProtectedData-price-handling

Le-Caignec:
This PR aims to fix the existing codebase on the develop branch to ensure the `workerpoolPrice` is properly included when calculating the `totalPrice`. This guarantees that the total cost of a task will not exceed the `maxPrice` provided by the user.

If a voucher covers part of the totalPrice, that amount is still considered as a cost for the requester.


PierreJeanjacquot:
I don't understand the whole logic.
Some of the method names are confusing (e.g., `fetchOrdersUnderMaxPrice` is a filtering method that doesn't fetch anything), so it doesn't help.
I think this is worth refactoring 

IMO, the logic should be:
- get cheapest app/dataset/workerpool orders given specified conditions (voucher or not)
- call `iexec.orders.estimateMatchOrders`
- check `(estimated.total - estimated.sponsored) <= maxPrice`

packages/sdk/tests/utils/mockAllForProcessProtectedData.ts
```diff
@@ -1,4 +1,5 @@
 import { jest } from '@jest/globals';
+import BN from 'bn.js'; // TODO use utils.BN from iexec and remove bn.js deps when https://github.com/iExecBlockchainComputing/iexec-sdk/pull/354 is released
```
PierreJeanjacquot, line 2: leaved this TODO

PierreJeanjacquot (approved):
LGTM

PR #426, opened 2025-04-14
Feature/update processprotecteddata

SeddikBellamine:
- Use promise.all 
- Add dataMaxPrice & appMaxPrice & workerpoolMaxPrice as parameters


packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
```diff
@@ -136,76 +146,111 @@ export const processProtectedData = async ({
     }
 
     vOnStatusUpdate({
-      title: 'FETCH_PROTECTED_DATA_ORDERBOOK',
```
gfournieriExec, line 139: Is it still usefull

packages/sdk/src/lib/types/coreTypes.ts
```diff
@@ -276,6 +276,7 @@ export type TransferResponse = {
 
 // ---------------------ProcessProtectedData Types------------------------------------
 export type ProcessProtectedDataStatuses =
+  | 'FETCH_ORDERS'
   | 'FETCH_PROTECTED_DATA_ORDERBOOK'
```
gfournieriExec, line 280: is it still usefull here ?

packages/sdk/tests/unit/dataProtectorCore/processProtectedData/processProtectedData.test.ts
```diff
@@ -184,10 +185,10 @@ describe('processProtectedData', () => {
       });
     });
 
-    describe('When maxPrice is not a positive number', () => {
+    describe('When DatamaxPrice is not a positive number', () => {
```
gfournieriExec, line 188:
```suggestion
    describe('When DataMaxPrice is not a positive number', () => {
```

gfournieriExec (approved):
LGTM thanks, just check why there is an issue with drone CI :) 
+ some minor comments

packages/sdk/src/lib/types/coreTypes.ts
SeddikBellamine: in the previous version we use the fetch orderbook for protected data & app & workerpool, but when we use promise.all all the fetches for orders is done in paralell so that's why i added this step to do the three fetches in one step

PR #436, opened 2025-04-30
Feature/detect user chain

SeddikBellamine:
**Multi-Chain Support for DataProtector SDK**

This PR adds multi-chain support to the IExecDataProtectorModule, enabling the SDK to work with networks beyond Bellecour, such as Arbitrum and Avalanche.

**Changes**
Introduced a `CHAIN_CONFIG` map in `config.ts `for per-chain addresses and URLs.

Added a getChainIdFromProvider utility to detect the connected chain from various provider types.

Refactored IExecDataProtectorModule to use an `async static create()` method that initializes configuration dynamically based on the chain ID.

Ensured compatibility with both Node.js and browser environments.

Preserved fallback to Bellecour when the chain ID is unknown or unsupported.

This refactor makes the SDK extensible and ready for deployment on multiple EVM-compatible blockchains.


packages/sdk/src/utils/getChainId.ts
```diff
@@ -0,0 +1,34 @@
...
+      const network = await provider.getNetwork();
+      return Number(network.chainId);
+    } else if ('request' in ethProvider) {
+      const provider = new BrowserProvider(ethProvider as Eip1193Provider);
+      const network = await provider.getNetwork();
+      return Number(network.chainId);
+    } else if ('getNetwork' in ethProvider) {
+      const network = await (ethProvider as AbstractProvider).getNetwork();
+      return Number(network.chainId);
+    }
```
PierreJeanjacquot, line 29: - We can avoid using `"getNetwork" in ethProvider` check by checking ethers class instance, which is more restrictive
- AbstractSigner does not implement `getNetwork`, we need to use the connected provider if it exists
- `'request' in ethProvider` to test the EIP1193 case is the less restrictive check, and could have false positive matches,  we should perform it last 

```suggestion
    } else if (ethProvider instanceof AbstractProvider) {
      const network = await ethProvider.getNetwork();
      return Number(network.chainId);
    } else if (ethProvider instanceof AbstractSigner) {
      const { provider } = ethProvider;
      if (!provider) {
        throw Error('Signer is not connected to a provider');
      }
      const network = await provider.getNetwork();
      return Number(network.chainId);
    } else if ('request' in ethProvider) {
      const provider = new BrowserProvider(ethProvider as Eip1193Provider);
      const network = await provider.getNetwork();
      return Number(network.chainId);
    }
```

PierreJeanjacquot (changes requested):
- see comments for changes
- check CI (code format issue)
- conventional commit is not required (but would be nice) at the feature branch level, when not used we must make sure to use squash merge feature branches with a conventional commit message

packages/sdk/src/lib/IExecDataProtectorModule.ts
```diff
@@ -30,51 +38,60 @@ abstract class IExecDataProtectorModule {
...
-  constructor(
-    ethProvider?:
-      | AbstractProvider
-      | AbstractSigner
-      | Eip1193Provider
-      | Web3SignerProvider
-      | string,
+  protected constructor() {
+    // Prevent direct instantiation; use create() instead
+  }
```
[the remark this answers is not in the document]
SeddikBellamine: fixed [1312aa6](https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/436/commits/1312aa6e4752c6410dbc6f8f06a418bb87a7fbc5)

packages/sdk/src/config/config.ts
```diff
@@ -1,34 +1,57 @@
...
-
-export const DEFAULT_SHARING_CONTRACT_ADDRESS =
-  '0x1390c3c6a545198809F1C7c5Dd2600ef74D60925'.toLowerCase();
-
-export const DEFAULT_SUBGRAPH_URL =
-  'https://thegraph.iex.ec/subgraphs/name/bellecour/dataprotector-v2';
-
-export const DEFAULT_DEBUG_SMS_URL = 'https://sms-debug.iex.ec';
-
 export const WORKERPOOL_ADDRESS = 'prod-v8-bellecour.main.pools.iexec.eth';
```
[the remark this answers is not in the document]
SeddikBellamine: https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/436/commits/1038ad5f8731ac06ebeea3696d73de0ac10ad61f

packages/sdk/src/lib/IExecDataProtectorModule.ts
```diff
@@ -30,51 +38,60 @@ abstract class IExecDataProtectorModule {
...
+  static async create(
+    ethProvider?: EthersCompatibleProvider,
     options?: DataProtectorConfigOptions
-  ) {
-    const ipfsGateway = options?.ipfsGateway || DEFAULT_IPFS_GATEWAY;
+  ): Promise<IExecDataProtectorModule> {
+    const instance = Object.create(this.prototype) as IExecDataProtectorModule;
+    const resolvedEthProvider = ethProvider || 'bellecour';
+    const chainId = await getChainIdFromProvider(resolvedEthProvider);
+    const config = CHAIN_CONFIG[chainId] || CHAIN_CONFIG[DEFAULT_CHAIN_ID];
```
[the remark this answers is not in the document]
SeddikBellamine: https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/436/commits/06c54c47754ac94b967a6fcd718649c07fec561a

packages/sdk/src/lib/IExecDataProtectorModule.ts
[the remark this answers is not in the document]
SeddikBellamine: i'm not sure this is a good idea as calling init in the operations require type of methods changed in specific method to async which is a breaking change

packages/sdk/tests/unit/constructor.test.ts
```diff
@@ -162,22 +168,19 @@ describe('IExecDataProtector()', () => {
...
   });
 
   describe('When instantiating SDK without a signer', () => {
     describe('When calling a write method', () => {
       it('should throw the corresponding exception', async () => {
-        // --- GIVEN
         const dataProtector = new IExecDataProtector();
-
-        // ---- WHEN / THEN
+        await dataProtector['init']();
```
PierreJeanjacquot, line 180: we should not need to call init manually here, `dataProtector.core.protectData` should do it

packages/sdk/src/config/config.ts
```diff
@@ -1,34 +1,63 @@
...
+    name: 'avalanche',
+    dataprotectorContractAddress: '0xYourAvalancheContractAddress',
+    sharingContractAddress: '0xYourAvalancheSharingAddress',
+    subgraphUrl:
+      'https://api.thegraph.com/subgraphs/name/your/avalanche-subgraph',
+    ipfsGateway: 'https://your-avalanche-ipfs-gateway',
+    ipfsNode: 'https://your-avalanche-ipfs-node',
+    smsDebugURL: 'https://your-avalanche-sms-url',
+    workerpoolAddress: 'WorkerpoolAddress',
+  },
```
PierreJeanjacquot, line 52: remove this until the whole stack deployment is ready on these chains

```suggestion
```

packages/sdk/src/lib/dataProtectorSharing/consumeProtectedData.ts
PierreJeanjacquot, line 1: Do not resolve `workerpoolAddress` here.
Instead, do it in the `init()` method with any other config resolution and inject it as `defaultWorkerpool` param.

packages/sdk/tests/unit/dataProtectorCore/processProtectedData/processProtectedData.test.ts
PierreJeanjacquot, line 1: With the default workerpool address resolution moved in the `init()` method, these changes in the test should be useless.

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
PierreJeanjacquot, line 1: Do not resolve `workerpoolAddress` here.
Instead, do it in the `init()` method with any other config resolution and inject it as `defaultWorkerpool` param.

packages/sdk/src/lib/IExecDataProtectorModule.ts
PierreJeanjacquot, line 1: It seems that a protected key `defaultWorkerpool` is missing in the class `IExecDataProtectorModule`, which forces doing the resolution in `processProtectedData` and `consumeProtectedData`; this should be avoided.

PierreJeanjacquot (changes requested):
Good job refactoring this :muscle:

I found another configuration resolution (default workerpool to use) that is not done at the right place and should be moved to the class initialization. I think we should address this point, sorry for missing this point on the first review :worried:

packages/sdk/tests/unit/constructor.test.ts
SeddikBellamine: the problem is that here it's unit tests and we don't execute them on local stack the stack, i propose we keep the init because it's unit test

SeddikBellamine: can't call the protect data or any other method

packages/sdk/src/lib/dataProtectorSharing/consumeProtectedData.ts
SeddikBellamine: add here https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/436/commits/d5c910509abc4fbfa29eaa3aa96de7454bffd6c1

packages/sdk/src/lib/dataProtectorCore/processProtectedData.ts
SeddikBellamine: https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/436/commits/d5c910509abc4fbfa29eaa3aa96de7454bffd6c1

packages/sdk/tests/unit/dataProtectorCore/processProtectedData/processProtectedData.test.ts
SeddikBellamine: https://github.com/iExecBlockchainComputing/dataprotector-sdk/pull/436/commits/ede37e4d1ad482cda31dbd816c55178021bcef24

PierreJeanjacquot, line 1: Whole changes in this file should be revertable

PierreJeanjacquot (commented):
see comments to remove remaining config resolution in processProtectedData
stopping review for now as the test worflow is bugged (subgraph build fails introduced in #437)

packages/sdk/src/lib/dataProtectorSharing/consumeProtectedData.ts
```diff
@@ -109,8 +113,9 @@ export const consumeProtectedData = async ({
     isDone: false,
   });
   try {
+    const { chainId } = await iexec.network.getNetwork();
     const workerpoolOrderbook = await iexec.orderbook.fetchWorkerpoolOrderbook({
-      workerpool: vWorkerpool || WORKERPOOL_ADDRESS,
+      workerpool: vWorkerpool || CHAIN_CONFIG[chainId].workerpoolAddress,
```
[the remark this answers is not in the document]
PierreJeanjacquot: f9c109ebff8cd12982942f6c72b8af6f59cd207d

PierreJeanjacquot (approved):
LGTM :)

PR #447, opened 2025-05-21
Fix/fix scripts subgraph package

Le-Caignec:


PierreJeanjacquot (approved):
LGTM

PierreJeanjacquot (commented):
custom start block seams to be missing, we need it to start indexing at the fork level

PierreJeanjacquot (changes requested):
Custom start block injection in the deployer is needed
For the sdk test purpose, we need to configure the subgraph deployer to start indexing from the BELLECOUR_FORK_BLOCK value to avoid a full subgraph resync which is not possible in the tests timeframe.

PierreJeanjacquot (approved):
LGTM thank you :)

PR #456, opened 2025-06-25
feat(sdk): add experimental networks

PierreJeanjacquot:
- Added `allowExperimentalNetworks` option (default `false`) to `IExecDataProtector`'s constructor and `getWeb3Provider()`. When `allowExperimentalNetworks: true` providers connected to experimental networks can be used
- Added `arbitrum-sepolia-testnet` as an experimental network


packages/sdk/src/config/config.ts
```diff
@@ -24,6 +25,30 @@ export const CHAIN_CONFIG: Record<ChainId, ChainConfig> = {
...
+    name: 'arbitrum-sepolia-testnet',
+    dataprotectorContractAddress: '0x2296daeDD3090750a80fFB2D0147669984909ED2',
+    sharingContractAddress: '0x2485Ed90d4566516298B7D01462df8d1A41E13AE',
+    subgraphUrl:
+      'https://thegraph.arbitrum-sepolia-testnet.iex.ec/api/subgraphs/id/5YjRPLtjS6GH6bB4yY55Qg4HzwtRGQ8TaHtGf9UBWWd',
+    ipfsGateway: 'https://ipfs-gateway.arbitrum-sepolia-testnet.iex.ec',
+    ipfsNode: 'https://ipfs-upload.arbitrum-sepolia-testnet.iex.ec',
+    smsDebugURL: 'https://sms.arbitrum-sepolia-testnet.iex.ec', // ⚠️ default SMS is a debug SMS
+    workerpoolAddress: '0x39c3cdd91a7f1c4ed59108a9da4e79de9a1c1b59',
+    isExperimental: true,
```
SeddikBellamine, line 39: what do you think of adding some kind of validation to the blockchain configurations? just a suggestion it's out of scope for this ticket i guess

SeddikBellamine (approved):
Good job, LGTM! the constructor tests are great but the other tests are passing because we run them on bellecour, shouldn't we consider in the future adding a test stack for the network that has been added and execute tests on all the existing networks configurations with wallets that have fake funds in them to simulate real transactions ?
it's there that we might find some new kind or errors related to the chain and txs fees

packages/sdk/src/config/config.ts
PierreJeanjacquot: yes it would be nice to validate user inputs

PierreJeanjacquot:
> Good job, LGTM! the constructor tests are great but the other tests are passing because we run them on bellecour, shouldn't we consider in the future adding a test stack for the network that has been added and execute tests on all the existing networks configurations with wallets that have fake funds in them to simulate real transactions ? it's there that we might find some new kind or errors related to the chain and txs fees

arbitrum-sepolia-testnet is experimental and may be removed anytime soon, so I don't think it is worth it on this network, but we may want to reconsider the question for non-experimental networks.

PR #457, opened 2025-06-30
fix: add check for nested empty JSON objects

Le-Caignec:


packages/subgraph/src/dataProtector.ts
```diff
@@ -111,7 +111,8 @@ function recursiveParse(
 
     if (entry.value.kind == JSONValueKind.OBJECT) {
       const object = entry.value.toObject();
-      for (let i = 0; i < object.entries.length; i++) {
```
Le-Caignec, line 114: This for loop is a mistake. We shouldn’t keep it, I think.

gfournieriExec: ok, I don't have the history of this

zguesmi: @Le-Caignec do you remember why was it added?

zguesmi (approved):

PierreJeanjacquot (approved):

PR #469, opened 2025-07-24
fix/dataProtector sharing fix ci

Le-Caignec:


Le-Caignec (commented):
### Deployment ID Issue:

Could you rename the existing committed deployment-id to chain-421614, or just remove it?

### Code Modification :

Could you modify this line in the deployment script: `const proxyAdminOwner = m.getAccount(0);` to use the ADMIN_PRIVATE_KEY environment variable instead? According to our workflow, the deployer should be the admin.

zguesmi (approved):

zguesmi (approved):

PR #472, opened 2025-07-28
fix: Init proxy contracts at deployment

Le-Caignec:


packages/sharing-smart-contract/contracts/DataProtectorSharing.sol
```diff
@@ -66,13 +66,15 @@ contract DataProtectorSharing is
         ADD_ONLY_APP_WHITELIST_REGISTRY = addOnlyAppWhitelistRegistry_;
     }
 
-    function initialize() public initializer {
+    function initialize(address admin) public initializer {
```
Le-Caignec, line 69:
```
function initialize(address admin) public initializer {
        __ERC721_init("iExec DataProtectorSharing", "iExecDataProtectorSharing");
        __ERC721Burnable_init();
        __AccessControl_init();
        _grantRole(DEFAULT_ADMIN_ROLE, admin);

        _iexecResultStorageProvider = "ipfs";
    }
 ```
 
 I think, it's better to do this. What do you think ?

